Issue #4, opened 2018-02-01
Allow user to supply Environment variables to the chart on the command line


hr1sh1kesh:
**Is this a BUG REPORT or FEATURE REQUEST?**:

**What happened**:
Environment variables that are passed on to Portworx specification should be allowed to be via the --set option while `helm install`

**What you expected to happen**:
`helm install --debug --set envVars=myENVVAR1=val1, myENVVAR2=val2 ./portworx-helm`
**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Container Orchestrator and version:
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

Issue #3, opened 2018-02-01
Enable deletion of Portworx Runc containers upon helm delete


hr1sh1kesh:
**Is this a BUG REPORT or FEATURE REQUEST?**:
FEATURE REQUEST
**What happened**:
Enable deletion of the PX runc containers upon `helm delete`.
**What you expected to happen**:
When a user runs `helm delete <release-name>` the Portworx runc containers that run on the nodes should also get deleted. 
**How to reproduce it (as minimally and precisely as possible)**:
run `helm delete <release-name>` the kubernetes pods and the applied spec would be deleted but the portworx runc containers which run on the nodes would still be running. 

**Anything else we need to know?**:

**Environment**:
- Container Orchestrator and version:
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

harsh-px:
@hr1sh1kesh is this fixed with the post and pre delete hooks?

harsh-px:
I verified this now and this appears to be implemented and working fine.

PR #5, opened 2018-02-02
Helm Chart for Installing Portworx on Kubernetes

### LINKED ISSUE #2: Develop a Helm chart which would ease the deployment of Portworx
**Is this a BUG REPORT or FEATURE REQUEST?**:
FEATURE REQUEST
**What happened**:
Develop a Helm chart which would ease the deployment of Portworx on kubernetes. 
**What you expected to happen**:
Should be able to deploy Portworx using the helm client. 

**How to reproduce it (as minimally and precisely as possible)**:

**Environment**:
- Container Orchestrator and version:
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

hr1sh1kesh:
Helm chart for installing Portworx on Kubernetes. 

Closes #2 , #3 , #4 

@harsh-px - Ready for review.


charts/px/templates/hooks/post-delete/px-postdelete-unlabelnode.yaml
```diff
@@ -0,0 +1,33 @@
...
+spec:
+  backoffLimit: 1
+  template:
+    spec:
+      serviceAccountName: tiller
+      restartPolicy: OnFailure
+      volumes:
+      - name: kubectl
+        hostPath:
+          path: /usr/bin/kubectl
```
harsh-px: We are assuming `/usr/bin/kubectl` is present on the host path in the system. What is the guarantee?

charts/px/templates/portworx-ds-zerostorage.yaml
```diff
@@ -0,0 +1,254 @@
...
+  
+  {{- else if eq (.Values.deploymentType | upper | lower) "docker" }}
+  updateStrategy:
+    type: OnDelete
+  
+  {{- end }}
+  template:
+    metadata:
+      labels:
+        app:  portworx-zerostorage
```
harsh-px: Let's deprecate the zerostorage. Moving forward, we don't need to install on the master. So the entire zerostorage spec isn't needed.

charts/px/templates/portworx-ds.yaml
```diff
@@ -0,0 +1,254 @@
...
+              {{- end }}
+
+              {{- if not (eq $managementInterface "none") }}
+              "-m",
+              "{{ $managementInterface }}",
+              {{- end }}
+
+
+              {{- if not (eq $lighthouseToken "none") }}
+              "-t",
```
harsh-px: remove lighthouseToken from all places. We aren't supporting it.

charts/px/values.yaml
```diff
@@ -0,0 +1,30 @@
...
+  imageType: none                         # 
+  imageVersion: 1.2.14                  # Version of the PX Image. 
+  openshiftInstall: false                 # Defaults to false for installing Portworx on Openshift .
+  isTargetOSCoreOS: false                 # Is your target OS CoreOS? Defaults to false.
+
+  etcdEndPoint:                         # The ETCD endpoint. Should be in the format etcd:http://<your-etcd-endpoint>:2379 
+                                        # the default value is empty since it requires to be explicity set using either the --set option of -f values.yaml.  
+  clusterName: mycluster                # This is the default. please change it to your cluster name.  
+  runOnMaster: false                    # Instructs Portworx to be run on the kubernetes master as well. Defaults to false.  
+  zeroStorage: false                    # true/false Instructs PX to run in Zero Storage mode.
```
harsh-px: remove zeroStorage from all places

charts/px/values.yaml
```diff
@@ -0,0 +1,30 @@
...
+  zeroStorage: false                    # true/false Instructs PX to run in Zero Storage mode. 
+  usefileSystemDrive: false             # true/false Instructs PX to use an unmounted Drive even if it has a filesystem.
+  usedrivesAndPartitions: false         # Defaults to false. Change to true and PX will use unmounted drives and partitions. 
+  secretType: none                      # Defaults to None, but can be AWS / KVDB / Vault.
+  drives: none                          # Defaults to use -A switch. 
+  dataInterface: none                   # Name of the interface <ethX>
+  managementInterface: none             # Name of the interface <ethX>
+  envVars: none                       # NOTE: This is a ":" seperated list of environment variables. For eg: MYENV1=myvalue1:MYENV2=myvalue2 
+    
+  lighthouse:
```
harsh-px: remove lighthouse. We will add it back once we add full support.

charts/px/README.md
```diff
@@ -0,0 +1,88 @@
...
+```bash
+$ helm delete my-release
+```
+
+The command removes all the Kubernetes components associated with the chart and deletes the release.
+
+## Configuration
+
+The following tables lists the configurable parameters of the Datadog chart and their default values.
+
```
harsh-px: helm chart needs to support the new stork=true parameter which installs stork.

hr1sh1kesh: Done. Should current default option for `stork` is `false` though

charts/px/templates/hooks/post-delete/px-postdelete-unlabelnode.yaml
hr1sh1kesh: We could move it to `/opt/pwx` and call the command? That would be a safe bet since this is a delete hook and we could assume `/opt/pwx` is going to be there. I would fix this upon your comment @harsh-px

harsh-px: My question is regarding the source location from where will you move it.

```
hostPath:
           path: /usr/bin/kubectl
```
means you are assuming that the host where this job runs will have a binary `/usr/bin/kubectl` at that location. What happens if it doesn't exist?

charts/px/README.md
harsh-px: had discussed this with @disrani-px  and we will have the default as false for now. Let's have it see some field time.

PR #47, opened 2018-07-18
Fixes for GKE install and Internal KVDB conditional check

### LINKED ISSUE #46: helm chart install on GKE does not install the PVC controller
**BUG REPORT **: When installing on GKE, helm install should detect GKE using the k8s version and also install the pvc controller. The spec generator at install.portworx.com currently uses the "gke" field in the kubectl version output to detect it's GKE and includes the pvc controller in the spec response.

**What happened**:
As the pvc controller does not install installed, the PVC creations will fail.

hr1sh1kesh:
Fixes the GKE install which requires the PVC Controller. 
Fixes the Internal KVDB check, where it allows for an empty etcdEndPoint when internalKVDB is set to true

Closes #46 

### DIFF
--- a/charts/portworx/templates/_helpers.tpl
+++ b/charts/portworx/templates/_helpers.tpl
@@ -70,52 +70,27 @@ release: {{ .Release.Name | quote }}
 {{- end -}}
 {{- end -}}
 
-{{- define "px.getk8sSchedulerImage" -}}
+{{- define "px.getk8sImages" -}}
 {{- if (.Values.customRegistryURL) -}}
     {{- if (eq "/" (.Values.customRegistryURL | regexFind "/")) -}}
-        {{ cat (trim .Values.customRegistryURL) "/kube-scheduler-amd64" | replace " " ""}}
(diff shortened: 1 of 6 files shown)


charts/portworx/templates/_helpers.tpl
```diff
@@ -26,6 +26,74 @@ release: {{ .Release.Name | quote }}
...
+{{- end -}}
+{{- end -}}
+
+{{- define "px.getStorkImage" -}}
+{{- if (.Values.customRegistryURL) -}}
+    {{- if (eq "/" (.Values.customRegistryURL | regexFind "/")) -}}
+        {{ cat (trim .Values.customRegistryURL) "/stork" | replace " " ""}}
+    {{- else -}}
+        {{- if .Values.openshiftInstall -}}
+            {{cat (trim .Values.customRegistryURL) "/portworx/stork" | replace " " ""}}
```
harsh-px: @hrish1kesh now we will just use the default stork image even for openshift. So you can remove this check.

PR #66, opened 2018-11-12
Remove portworx namespace creation and px-role from helm chart

adityadani:
1. We cannot create portworx namespace in the helm chart, as the helm chart installation fails
    if the namespace already exists.
2. If the `portworx` namespace does not exist we cannot create the px-role for that namespace.

The installation process for PX + IBM Key Protect would be like this

1. Create portworx namespace
2. Create the px-role and the role bindings for a non-existent service account (px-account) -> k8s does not complain on this
3. Create the px-ibm k8s secret
4. Install the helm chart

With this flow, PX when installed will automatically get authenticated with IBM KP.

Closes #

### DIFF
--- a/charts/portworx/templates/portworx-rbac-config.yaml
+++ b/charts/portworx/templates/portworx-rbac-config.yaml
@@ -42,34 +42,3 @@ roleRef:
   kind: ClusterRole
   name: node-get-put-list-role
   apiGroup: rbac.authorization.k8s.io
-
----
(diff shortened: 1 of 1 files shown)


harsh-px:
Aren't steps 1, 2 and 3 needed only when you need to create an encrypted volume?

adityadani:
Yes these steps are only required for IBM KP integration.
And only those related docs, will have it.

PR #69, opened 2019-01-22
Fixing mounts and adding install annotations

zoxpx:
* fixing helm-chart mounts (adding journalmount* for internal log-proxy, also fixing PKS)
* adding annotations to capture install-vars (useful for upgrades when helm-home lost)

Signed-off-by: Zoran Rajic <zox@portworx.com>

Helm had mount-points discrepancy compared to install.portworx.com
Also, will be very useful to capture install-vars as ds/portworx annotations  (not perfect, but enough to be able to manually reconstruct the install/upgrade).

-

### DIFF
--- a/charts/portworx/templates/portworx-ds.yaml
+++ b/charts/portworx/templates/portworx-ds.yaml
@@ -32,6 +32,9 @@ kind: DaemonSet
 metadata:
   name: portworx
   namespace: kube-system
+  annotations:
+    portworx.com/install-source: helm/{{.Release.Service}}-r{{.Release.Revision}}
+    portworx.com/helm-vars: chart="{{.Chart.Name}}-{{.Chart.Version}}"{{range $k, $v := .Values }}{{if $v}},{{ $k }}="{{ $v }}"{{end}}{{end}}
 spec:
   minReadySeconds: 0
   updateStrategy:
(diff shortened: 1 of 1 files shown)


zoxpx:
**NOTE**: here's how the new annotations will look like (should be useful to reconstruct upgrade/install)

```yaml
metadata:
  annotations:
    portworx.com/helm-vars: chart="portworx-1.0.0",clusterName="k8s-cri-r4",consul="map[token:none]",dataInterface="eth1",deploymentType="oci",drives="/dev/sdb",envVars="none",etcd="map[ca:none
      cert:none certPath:none credentials:none:none key:none]",etcdEndPoint="etcd:http://70.0.0.65:2379",imageType="none",imageVersion="1.7.5",lighthouseVersion="1.5.0",managementInterface="eth1",secretType="none",serviceAccount="map[hook:map[create:true
      name:<nil>]]",storkVersion="1.2.0"
    portworx.com/install-source: helm/Tiller-r4
```

zoxpx:
Thanks @hrishipx -- I tested the changes using my IKS-like Kubernetes-CRI/Containerd (tests went fine).  Also did a number of tests with `helm install --dry-run..` to generate the YAML, and double-checked the paths and mounts against install.portworx.com -- so all in all, I think this should be "good enough" for PKS tests.

PR #70, opened 2019-01-22
Updates to few RBAC changes, defaulting secrets to k8s and values.yam…

hrishipx:
Signed off by : @hrishipx 

- Updates the RBAC changes for stork pods. 
- Updates values.yaml with the latest Portworx version. 
- Defaults the secrets provider as k8s (similar to how it is on the spec generator)

Closes #

### DIFF
--- a/charts/portworx/templates/portworx-k8s-secrets.yaml
+++ b/charts/portworx/templates/portworx-k8s-secrets.yaml
@@ -7,7 +7,7 @@ metadata:
   name: portworx
 ---
 kind: Role
-apiVersion: rbac.authorization.k8s.io/v1
+apiVersion: {{ template "rbac.apiVersion" . }}
 metadata:
   name: px-role
   namespace: portworx
@@ -17,7 +17,7 @@ rules:
   verbs: ["get", "list", "create", "update", "patch"]
 ---
 kind: RoleBinding
-apiVersion: rbac.authorization.k8s.io/v1
+apiVersion: {{ template "rbac.apiVersion" . }}
 metadata:
   name: px-role-binding
   namespace: portworx
--- a/charts/portworx/templates/portworx-k8s-secrets.yaml
+++ b/charts/portworx/templates/portworx-k8s-secrets.yaml
@@ -1,0 +1,32 @@
+{{- $secretType := .Values.secretType | default "k8s" }}
+
+{{- if eq $secretType "k8s" }}
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: portworx
+---
+kind: Role
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: px-role
+  namespace: portworx
+rules:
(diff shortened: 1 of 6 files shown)


charts/portworx/templates/portworx-k8s-secrets.yaml
```diff
@@ -0,0 +1,32 @@
...
+metadata:
+  name: px-role
+  namespace: portworx
+rules:
+- apiGroups: [""]
+  resources: ["secrets"]
+  verbs: ["get", "list", "create", "update", "patch"]
+---
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
```
zoxpx: Don't we have a global flag for the RBAC version?
 (afaik RBAC version changes depending on Kubernetes version)

charts/portworx/values.yaml
```diff
@@ -13,20 +13,20 @@ etcdEndPoint:                         # The ETCD endpoint. Should be in the form
 clusterName: mycluster                # This is the default. please change it to your cluster name.
 usefileSystemDrive: false             # true/false Instructs PX to use an unmounted Drive even if it has a filesystem.
 usedrivesAndPartitions: false         # Defaults to false. Change to true and PX will use unmounted drives and partitions.
-secretType: none                      # Defaults to None, but can be kvdb/k8s/aws-kms/vault/ibm-kp
+secretType: k8s                      # Defaults to None, but can be kvdb/k8s/aws-kms/vault/ibm-kp
```
zoxpx: NOTE: Comment is incorrect (Default is no longer none).

Please also ensure we are keeping the default identical to the defaults on http://install.portworx.com

charts/portworx/values.yaml
```diff
@@ -13,20 +13,20 @@ etcdEndPoint:                         # The ETCD endpoint. Should be in the form
...
 stork: true                           # Use Stork https://docs.portworx.com/scheduler/kubernetes/stork.html for hyperconvergence.
-storkVersion: 1.2.0
+storkVersion: 2.0.1
 
 customRegistryURL:
 registrySecret:
 
 lighthouse: true
-lighthouseVersion: 1.5.0
+lighthouseVersion: 2.0.1
```
zoxpx: Note that lighthouse on https://install.portworx.com/2.0/ is indeed 2.0.1, but it also uses an additional POD (portworx/lh-config-sync:0.2), which here we don't ask the version for.
Please make sure helm-YAML will use 2nd POD (also please let's have QA validate LH-install w/ this change)

charts/portworx/templates/portworx-k8s-secrets.yaml
hrishipx: Good catch @zoxpx . I have used the global template for it now.

charts/portworx/values.yaml
hrishipx: Updated the comments

PR #73, opened 2019-01-29
Bump up the Image versions after new release

### LINKED ISSUE #72: Update default values to more recent releases of Portworx
**Is this a BUG REPORT or FEATURE REQUEST?**: minor update

**What happened**: versions listed in values.yaml file for Portworx, Stork and Lighthouse should be update to new releases

**What you expected to happen**: changes in default values in values.yaml file

**How to reproduce it (as minimally and precisely as possible)**: this a not a bug, so no need to reproduce

**Environment**:
- Container Orchestrator and version:
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

hrishipx:
Signed off by : @hrishipx 

Update the latest versions of Portworx. 

Closes #72 

### DIFF
--- a/charts/portworx/templates/portworx-service.yaml
+++ b/charts/portworx/templates/portworx-service.yaml
@@ -8,8 +8,20 @@ metadata:
 spec:
   selector:
     name: portworx
+  type: NodePort
   ports:
     - name: px-api
       protocol: TCP
       port: 9001
       targetPort: 9001
+      nodePort: 30901
+    - name: px-kvdb
+      protocol: TCP
+      port: 9019
+      targetPort: 9019
+      nodePort: 30919
+    - name: px-sdk
+      protocol: TCP
+      port: 9021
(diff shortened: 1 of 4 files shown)


charts/portworx/templates/portworx-service.yaml
```diff
@@ -8,8 +8,20 @@ metadata:
 spec:
   selector:
     name: portworx
+  type: NodePort
```
zoxpx: Oh.. was this broken before, since we missed NodePort?

hrishipx: @zoxpx no if we dont specify NodePort, it will be defaulted to ClusterIP, and currently with all communication to the portworx-service is internal within the k8s cluster it should work. shouldnt be broken.

PR #84, opened 2019-03-05
PWX-7940: Change LH Role to a ClusterRole and Binding to a ClusterBinding. Also fix the permissions for stork api's

pault84:
Signed-off-by: Paul <paul@portworx.com>

In 2.0 we added stork-connector and px-central to lighthouse.
This requires the lighthouse pod to be able to list cluster objects.

For this we had to change to a ClusterRole and Binding.

I also added the parameter required to tell lighthouse we're running on kubernetes. 
This is so lighthouse knows it should ask for kubeConfigs so we can do migrations.

Closes #PWX-7940

### DIFF
--- a/charts/portworx/templates/portworx-lighthouse.yaml
+++ b/charts/portworx/templates/portworx-lighthouse.yaml
@@ -5,11 +5,11 @@ metadata:
   name: px-lh-account
   namespace: kube-system
 ---
-kind: Role
+kind: ClusterRole
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
-   name: px-lh-role
-   namespace: kube-system
+  name: px-lh-role
+  namespace: kube-system
 rules:
   - apiGroups: [""]
     resources: ["secrets"]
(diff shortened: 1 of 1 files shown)


pault84:
I'll need to make similar changes to the IBM and Rancher charts.

charts/portworx/templates/portworx-lighthouse.yaml
```diff
@@ -21,20 +21,20 @@ rules:
     resources: ["nodes", "services"]
     verbs: ["get", "list", "watch"]
   - apiGroups: ["stork.libopenstorage.org"]
-    resources: ["clusterpairs"]
-    verbs: ["get", "list", "create", "update", "delete"]  
+    resources: ["clusterpairs","migrations"]
```
harsh-px: stork at install.portworx.com also has `groupvolumesnapshots. Can you also add that here?

pault84: yessir

pault84: Any other sections that might have drastically changed?
I'll update them but feature owners should maintain their pieces IMO.

pault84:
@harsh-px this look ok now?

PR #85, opened 2019-03-07
Update to latest releases and specs

pault84:
Signed-off-by: Paul <paul@portworx.com>

Moves the helm charts to the latest versions.
Adds monitoring if desired.

We're currently a bit behind on versions and we didn't have monitoring present yet.
Rancher and IKS are even further behind.

### DIFF
--- a/charts/portworx/templates/portworx-monitor.yaml
+++ b/charts/portworx/templates/portworx-monitor.yaml
@@ -1,4 +1,4 @@
-{{- if (.Values.pxCentral) and (eq .Values.pxCentral true) -}}
+{{- if (.Values.monitoring) and (eq .Values.monitoring true) -}}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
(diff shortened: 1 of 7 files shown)


charts/portworx/templates/portworx-monitor.yaml
```diff
@@ -0,0 +1,504 @@
...
+  ports:
+    - name: web
+      port: 9093
+      protocol: TCP
+      targetPort: 9093
+  selector:
+    alertmanager: portworx
+---
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
```
harsh-px: wouldn't this cause the helm install to fail since it won't find the kind?

pault84: Pre-requisites in doc i'll add them there.

charts/portworx/values.yaml
```diff
@@ -26,9 +26,10 @@ customRegistryURL:
 registrySecret:
 
 lighthouse: true
-lighthouseVersion: 2.0.1
+lighthouseVersion: 2.0.2
 lighthouseSyncVersion: 0.3
 lighthouseStorkConnectorVersion: 0.1
+pxCentral: false
```
harsh-px: let's use the same param names as install.portworx.com. So let's use `monitoring`.

charts/portworx/templates/portworx-monitor.yaml
pault84: Or should i go straight to the newest way with lighthouse deploying the CRD's ?

PR #87, opened 2019-03-18
Fix client-key.pem not found error

tux-00:
When using specific etcd certificates, portworx container doesn't found the _client-key.pem_ (specified dynamically as an argument in templates/portworx-ds.yaml) because the file is always mounted as _client-key.key_ :
```
            {{- if ne $etcdKey "none" }}
            - key: client-key.pem
              path: client-key.key
            {{- end -}}
```

Command ran to install the chart:
```
helm install --debug --name portworx \
--set etcdEndPoint=etcd:https://X.X.X.X:2379,clusterName=$(uuidgen) \
--set usedrivesAndPartitions=true \
--set drives=/dev/vda9 \
--set etcd.certPath=/etc/pwx/etcdcerts \
--set etcd.ca=/etc/pwx/etcdcerts/ca.pem \
--set etcd.cert=/etc/pwx/etcdcerts/client.pem \
--set etcd.key=/etc/pwx/etcdcerts/client-key.pem .
```

No one.

### DIFF
--- a/charts/portworx/templates/portworx-ds.yaml
+++ b/charts/portworx/templates/portworx-ds.yaml
@@ -324,7 +324,7 @@ spec:
             {{- end -}}
             {{- if ne $etcdKey "none" }}
             - key: client-key.pem
-              path: client-key.key
+              path: client-key.pem
             {{- end -}}
           {{- end}}
         - name: dockersock


harsh-px:
Closing as stale

PR #97, opened 2019-07-22
[STORK] Affinity and Tolerations options on Deployments

anthonyhaussman:
Remove the nodeAffinity rule from stork/stork-scheduler deployments, Stork can run on any node, it talks to PX over the service IP.

Also, add the possibility to add tolerations on stork/storck-scheduler deployments. We use the same tolerations than PX deployments.


matanbaru:
+1

harsh-px:
#122 should handle this

PR #98, opened 2019-07-25
[PWX] Add replicas value / Fix key DaemonSet

anthonyhaussman:
Fix key localization into portworx DaemonSet and add a value to configure replicas.


harsh-px:
#126 will add the replicas value. #87 fixes the .pem value.

Hence closing this one.

PR #100, opened 2019-07-26
Updated versions for all chart components

trierra:

Updated versions for all chart components

### DIFF
--- a/charts/portworx/values.yaml
+++ b/charts/portworx/values.yaml
@@ -2,7 +2,7 @@
 
 deploymentType: oci                     # accepts "oci" or "docker"
 imageType: none                         #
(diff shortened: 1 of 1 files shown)


charts/portworx/values.yaml
```diff
@@ -2,7 +2,7 @@
 
 deploymentType: oci                     # accepts "oci" or "docker"
 imageType: none                         #
-imageVersion: 2.0.3.6                     # Version of the PX Image.
```
harsh-px: The README.md will also need changes. Currently it says default is 2.0.3.4.

harsh-px: I'm going to merge this now. Update the readme in a subsequent commit.

PR #101, opened 2019-08-09
PSG-98: Add azure env vars from secret for AKS installations

trierra:

To support AKS installations

Closes #PSG-98

updated oci-mon to v2.1.2 since 2.0.3.6 did not have azure support

### DIFF
--- a/charts/portworx/README.md
+++ b/charts/portworx/README.md
@@ -35,10 +35,11 @@ The following tables lists the configurable parameters of the Portworx chart and
 |             Parameter       |            Description             |                    Default                |
 |-----------------------------|------------------------------------|-------------------------------------------|
 | `deploymentType`            | The deployment type. Can be either docker/oci   | `oci`                 |
-| `imageVersion`              | The image tag to pull              | `2.0.3.4`                                  |
+| `imageVersion`              | The image tag to pull              | `2.1.3`                                  |
 | `openshiftInstall`               | Installing on Openshift? | `false`                               |
 | `pksInstall`               | Installing on Pivotal Container service? | `false`                               |
-| `AKSorEKSInstall`               | Installing on AKS(Azure Kubernetes service) or EKS (Amazon Elastic Container service) | `false`                               |
+| `EKSInstall`               | Installing EKS (Amazon Elastic Container service) | `false`                               |
+| `AKSInstall`               | Installing on AKS (Azure Kubernetes service) | `false`                               |
 | `etcdEndPoint`          | (REQUIRED) ETCD endpoint for PX to function properly in the form "etcd:http://<your-etcd-endpoint>". Multiple Urls should be semi-colon seperated example: etcd:http://<your-etcd-endpoint1>;etcd:http://<your-etcd-endpoint2>  | `etcd:http://<your-etcd-endpoint>`                    |
 | `clusterName`           | Portworx Cluster Name  | `mycluster`                                     |
 | `usefileSystemDrive`      | Should Portworx use an unmounted drive even with a filesystem ? | `false`                |
@@ -65,6 +66,12 @@ The following tables lists the configurable parameters of the Portworx chart and
 
 
 Specify each parameter using the `--set key=value[,key=value]` argument to `helm install`. For example,
+
+## Cloud installs
+
+#### Installing on AKS 
+
+Details are [here](https://docs.portworx.com/portworx-install-with-kubernetes/cloud/azure/aks/2-deploy-px/).
 
 > **Tip**: In this case the chart is located at `./helm/charts/portworx`, do change it as per your setup.
 ```
--- a/charts/portworx/templates/portworx-ds.yaml
+++ b/charts/portworx/templates/portworx-ds.yaml
@@ -49,7 +49,7 @@ spec:
         # {{- include "px.labels" . | indent 8 }}
     spec:
       tolerations:
-{{ toYaml .Values.tolerations | indent 8 }} 
+{{ toYaml .Values.tolerations | indent 8 }}
       affinity:
         nodeAffinity:
           requiredDuringSchedulingIgnoredDuringExecution:
@@ -82,18 +82,20 @@ spec:
           {{- with .Values }}
             [
               {{- if eq $drives "none" }}
-                {{- if eq $usedrivesAndPartitions true }}
-              "-A",
-                {{- else }}
-              "-a",
-                {{- end -}}
-                {{- if or $usefileSystemDrive $deployEnvironmentIKS }}
-              "-f",
-                {{- end }}
-              {{- else }}
+                  {{- if .AKSInstall -}}
+                   "-s", "type=Premium_LRS,size=150",
+                  {{- else if eq $usedrivesAndPartitions true }}
+                      "-A",
+                  {{- else }}
+                     "-a",
+                  {{- end -}}
+                 {{- if or $usefileSystemDrive $deployEnvironmentIKS }}
+                      "-f",
+                  {{- end }}
+              {{- else -}}
                 {{- $driveNames := $drives | split ";" }}
                 {{- range $index, $name := $driveNames }}
-              "-s", "{{ $name }}",
+                  "-s", "{{ $name }}",
                 {{- end -}}
               {{- end -}}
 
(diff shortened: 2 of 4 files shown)


charts/portworx/README.md
```diff
@@ -35,16 +35,18 @@ The following tables lists the configurable parameters of the Portworx chart and
 |             Parameter       |            Description             |                    Default                |
 |-----------------------------|------------------------------------|-------------------------------------------|
 | `deploymentType`            | The deployment type. Can be either docker/oci   | `oci`                 |
-| `imageVersion`              | The image tag to pull              | `2.0.3.4`                                  |
+| `imageVersion`              | The image tag to pull              | `2.1.2`                                  |
```
harsh-px: sorry this needs to be 2.1.3 now :)

charts/portworx/README.md
```diff
@@ -35,16 +35,18 @@ The following tables lists the configurable parameters of the Portworx chart and
...
 | `etcdEndPoint`          | (REQUIRED) ETCD endpoint for PX to function properly in the form "etcd:http://<your-etcd-endpoint>". Multiple Urls should be semi-colon seperated example: etcd:http://<your-etcd-endpoint1>;etcd:http://<your-etcd-endpoint2>  | `etcd:http://<your-etcd-endpoint>`                    |
 | `clusterName`           | Portworx Cluster Name  | `mycluster`                                     |
 | `usefileSystemDrive`      | Should Portworx use an unmounted drive even with a filesystem ? | `false`                |
 | `usedrivesAndPartitions`  | Should Portworx use the drives as well as partitions on the disk ? | `false`             |
 | `secretType`      | Secrets store to be used can be AWS KMS/KVDB/Vault/K8s/IBM Key Protect          | `none`                                    |
 | `drives` | Semi-colon seperated list of drives to be used for storage (example: "/dev/sda;/dev/sdb")           | `none`                                   |
+| `defaultDrives` | Defaults for AKS installation if drives are not provided           | `type=Premium_LRS,size=150`                                   |
```
harsh-px: `defaultDrives` doesn't need to be user facing. This will confuse non-AKS users.

Just let the AKSInstall drive the logic and internally we pick the defaults.

charts/portworx/values.yaml
```diff
@@ -2,18 +2,20 @@
 
 deploymentType: oci                     # accepts "oci" or "docker"
 imageType: none                         #
-imageVersion: 2.0.3.8                   # Version of the PX Image.
+imageVersion: 2.1.2                    # Version of the PX Image.
```
harsh-px: 2.1.3

Also change default stork to 2.2.4

charts/portworx/README.md
```diff
@@ -35,16 +35,18 @@ The following tables lists the configurable parameters of the Portworx chart and
 |             Parameter       |            Description             |                    Default                |
 |-----------------------------|------------------------------------|-------------------------------------------|
 | `deploymentType`            | The deployment type. Can be either docker/oci   | `oci`                 |
-| `imageVersion`              | The image tag to pull              | `2.0.3.4`                                  |
+| `imageVersion`              | The image tag to pull              | `2.1.2`                                  |
 | `openshiftInstall`               | Installing on Openshift? | `false`                               |
 | `pksInstall`               | Installing on Pivotal Container service? | `false`                               |
-| `AKSorEKSInstall`               | Installing on AKS(Azure Kubernetes service) or EKS (Amazon Elastic Container service) | `false`                               |
+| `EKSInstall`               | Installing EKS (Amazon Elastic Container service) | `false`                               |
+| `AKSInstall`               | Installing on AKS(Azure Kubernetes service) | `false`                               |
```
piyush-nimbalkar: nit: space after AKS `AKS (Azure...`

charts/portworx/templates/portworx-ds.yaml
```diff
@@ -82,18 +83,25 @@ spec:
...
-                {{- if eq $usedrivesAndPartitions true }}
-              "-A",
-                {{- else }}
-              "-a",
-                {{- end -}}
-                {{- if or $usefileSystemDrive $deployEnvironmentIKS }}
-              "-f",
-                {{- end }}
-              {{- else }}
+                {{- if .AKSInstall -}}
```
piyush-nimbalkar: I don't think we should expose default drives to users. This should be in the template. For now we can assume there is only one default drive.
```
              {{- if eq $drives "none" }}
                {{- if .AKSInstall -}}
                   "-s", "type=Premium_LRS,size=150",
                {{- else -}}
                  {{- if eq $usedrivesAndPartitions true }}
                      "-A",
                  {{- else }}
                     "-a",
                  {{- end -}}
                 {{- if or $usefileSystemDrive $deployEnvironmentIKS }}
                      "-f",
                  {{- end }}
                {{ end }}
              {{- else -}}
                {{- $driveNames := $drives | split ";" }}
                {{- range $index, $name := $driveNames }}
                  "-s", "{{ $name }}",
                {{- end -}}
              {{- end -}}
```

In future if we have defaults for EKS or other cloud providers, we can add another `else if` in that
```
              {{- if eq $drives "none" }}
                {{- if .AKSInstall -}}
                   "-s", "type=Premium_LRS,size=150",
                {{- else if .EKSInstall -}}
                   "-s", "type=gp2,size=150",
                {{- else -}}
                  {{- if eq $usedrivesAndPartitions true }}
                      "-A",
                  {{- else }}
                     "-a",
                  {{- end -}}
                 {{- if or $usefileSystemDrive $deployEnvironmentIKS }}
                      "-f",
                  {{- end }}
                {{ end }}
              {{- else -}}
                {{- $driveNames := $drives | split ";" }}
                {{- range $index, $name := $driveNames }}
                  "-s", "{{ $name }}",
                {{- end -}}
              {{- end -}}
```

charts/portworx/values.yaml
piyush-nimbalkar: @harsh-px Should we change this to 2.1.3?

harsh-px: Yes

charts/portworx/templates/portworx-ds.yaml
```diff
@@ -161,6 +169,23 @@ spec:
              ]
            {{- end }}
           env:
+          {{ if .Values.AKSInstall }}
+            - name: AZURE_CLIENT_SECRET
+              valueFrom:
+                secretKeyRef:
+                  name: px-azure
```
satchpx: how would a user know to create a secret `px-azure` before trying to do a `helm install`?

charts/portworx/README.md
trierra: makes sense

charts/portworx/templates/portworx-ds.yaml
piyush-nimbalkar: https://docs.portworx.com/portworx-install-with-kubernetes/cloud/azure/aks/2-deploy-px/#install

charts/portworx/templates/portworx-ds.yaml
satchpx: To add to this, we should also make sure, we have support to specify metadata drive if `internalKVDB: true`.

charts/portworx/templates/portworx-ds.yaml
trierra: it's explained here https://docs.portworx.com/portworx-install-with-kubernetes/cloud/azure/aks/2-deploy-px/#create-a-secret-to-give-portworx-access-to-azure-apis

trierra:
done

charts/portworx/templates/portworx-ds.yaml
satchpx: Right. My point was, can we also add a note in our helm readme? We cannot expect someone to have landed on our helm docs from docs.portworx.com

charts/portworx/templates/portworx-ds.yaml
```diff
@@ -82,18 +82,23 @@ spec:
...
-              "-a",
-                {{- end -}}
-                {{- if or $usefileSystemDrive $deployEnvironmentIKS }}
-              "-f",
-                {{- end }}
-              {{- else }}
+                  {{- if .AKSInstall -}}
+                   "-s", "type=Premium_LRS,size=150",
+                  {{ end }}
+                {{- else -}}
```
piyush-nimbalkar: This `else` doesn't seem correct.

charts/portworx/templates/portworx-ds.yaml
harsh-px: @trierra Let's add a section in the README.

## Cloud installs

### Installing on AKS

Add details about how to create the secret here.

PR #111, opened 2019-12-06
PSG-134 Switch apiVersion to apps/v1 for DaemonSet and Deployment

rodrigc:

This is needed to have helm install work on Kubernetes 1.16

PSG-134


rodrigc:
I tried the following on a Kubernetes 1.11 cluster, and also on a Kubernetes 1.16 cluster:

```
git clone -b PSG-134 https://github.com/rodrigc/helm

helm install --debug  --set etcdEndPoint=etcd:http://70.0.5.211:2379,clusterName=$(uuidgen) portworx-craig ./helm/charts/portworx/
```

and was able to successfully install with the helm chart

PR #117, opened 2020-02-06
PWX-11024 px-sdk needs to be on port 9020

rodrigc:
px-sdk needs to be on port 9020.   This matches what is generated by the spec generator

### DIFF
--- a/charts/portworx/templates/portworx-service.yaml
+++ b/charts/portworx/templates/portworx-service.yaml
@@ -8,7 +8,7 @@ metadata:
 spec:
   selector:
     name: portworx
-  type: NodePort
+  type: ClusterIP
   ports:
     - name: px-api
       protocol: TCP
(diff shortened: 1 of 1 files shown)


charts/portworx/templates/portworx-service.yaml
```diff
@@ -19,6 +19,10 @@ spec:
       port: 9019
       targetPort: 9019
     - name: px-sdk
+      protocol: TCP
```
harsh-px: let's also remove `type: NodePort`. 

@trierra I just noticed these specs are missing the portworx-api daemonset and the k8s service. Can you add that in a later change?

rodrigc: @harsh if I look at: https://install.portworx.com/2.3?mc=false&kbver=1.14.7&b=true&f=true&j=auto&c=px-cluster-a29d12d9-872e-4d54-b87a-b4a74a29e560&stork=true&lh=true&st=k8s , we have:

`type: ClusterIP`, is that what I should change it to?

rodrigc:
I got one successful test run: https://jenkins.portworx.co/job/Torpedo/view/Torpedo%20NextPX/job/tp-nextpx-helm/17/

PR #121, opened 2020-04-28
Add tolerations from `values.yaml` to the templates

fernandodeperto:
When the chart is deployed to a cluster that contains nodes with taints, some of the deployments fail for not having the necessary `tolerations` statement.

This PR passes the `tolerations` value through to all the templates that have a `nodeSelector` statement.

### DIFF
--- a/charts/portworx/templates/portworx-controller.yaml
+++ b/charts/portworx/templates/portworx-controller.yaml
@@ -120,6 +120,8 @@ spec:
           requests:
             cpu: 200m
       hostNetwork: true
+      tolerations:
+{{ toYaml .Values.tolerations | indent 8 }}
       affinity:
         nodeAffinity:
           requiredDuringSchedulingIgnoredDuringExecution:
--- a/charts/portworx/templates/portworx-storageclasses.yaml
+++ b/charts/portworx/templates/portworx-storageclasses.yaml
@@ -1,56 +1,9 @@
+---
 kind: StorageClass
-apiVersion: storage.k8s.io/v1
+apiVersion: storage.k8s.io/v1beta1
 metadata:
-  name: portworx-db-sc
+  name: portworx-sc
 provisioner: kubernetes.io/portworx-volume
 parameters:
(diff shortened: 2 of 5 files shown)


charts/portworx/templates/portworx-controller.yaml
```diff
@@ -120,6 +120,8 @@ spec:
           requests:
             cpu: 200m
       hostNetwork: true
+      tolerations:
+{{ toYaml .Values.tolerations | indent 8 }}
```
harsh-px, line 124: wouldn't this cause an invalid yaml template if tolerations are empty in the values.yaml. In other words, this will get rendered as

```
  hostNetwork: true
  tolerations:
  affintiy:
    nodeAffinity:
    ...
```

charts/portworx/templates/portworx-storageclasses.yaml
```diff
@@ -1,56 +1,9 @@
-kind: StorageClass
```
harsh-px, line 1: was this intentional? This removes the default storage classes that the chart installs.

charts/portworx/templates/portworx-controller.yaml
fernandodeperto: Good catch. In our setup we always need tolerations so I didn't test without it. 

To keep the pattern, I copied the code over from the main daemon set, here:
https://github.com/portworx/helm/blob/master/charts/portworx/templates/portworx-ds.yaml#L57

So that probably needs improved as well. I'll update the PR.

charts/portworx/templates/portworx-storageclasses.yaml
fernandodeperto: No, that was meant to stay withing my fork of the repo. I'll update the PR to remove that.

fernandodeperto:
The best course of action here seems to recreate the PR.

fernandodeperto:
New PR here:
https://github.com/portworx/helm/pull/122

PR #122, opened 2020-04-30
Fix tolerations

fernandodeperto:
When the chart is deployed to a cluster that contains nodes with taints, some of the deployments fail for not having the necessary `tolerations` statement.

This PR passes the `tolerations` value through to all the templates that have a `nodeSelector` statement.


fernandodeperto:
I rebased the branch to include the changes made by #123.

PR #123, opened 2020-05-01
upgrade px to 2.5.1

trierra:
added portworx-api service

Closes #

### DIFF
--- a/charts/portworx/templates/portworx-ds.yaml
+++ b/charts/portworx/templates/portworx-ds.yaml
@@ -425,3 +425,50 @@ spec:
           hostPath:
             path: /proc
         {{- end }}
+---
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: portworx-api
+  namespace: kube-system
+  labels:
+    name: portworx-api
+spec:
+  selector:
+    matchLabels:
+      name: portworx-api
+  minReadySeconds: 0
+  updateStrategy:
+    type: RollingUpdate
+    rollingUpdate:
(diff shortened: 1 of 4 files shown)


charts/portworx/templates/portworx-ds.yaml
```diff
@@ -425,3 +425,50 @@ spec:
...
+                    operator: NotIn
+                    values:
+                      - "false"
+                  - key: node-role.kubernetes.io/master
+                    operator: DoesNotExist
+      hostNetwork: true
+      hostPID: false
+      containers:
+        - name: portworx-api
+          image: k8s.gcr.io/pause:3.1
```
harsh-px, line 465: can't use `k8s.gcr.io/pause:3.1` directly here since we need to support custom registries.  You can view other manifests in this chart as examples on how the image field gets populated.

charts/portworx/values.yaml
```diff
@@ -21,15 +21,15 @@ managementInterface: none             # Name of the interface <ethX>
...
 customRegistryURL:
 registrySecret:
 
 lighthouse: true
-lighthouseVersion: 2.0.5
-lighthouseSyncVersion: 2.0.5
-lighthouseStorkConnectorVersion: 2.0.5
+lighthouseVersion: 2.0.6
+lighthouseSyncVersion: 2.0.6
+lighthouseStorkConnectorVersion: 2.0.6
```
harsh-px, line 32: all of these are 2.0.7 now. I compared with spec returned by current https://install.portworx.com/?comp=lighthouse

PR #128, opened 2020-05-29
Improved secret management.

sean-mcgimpsey:
Previously, the cluster-wide-secret-key of the px-vol-encryption secret was regenerated on every helm upgrade. Potentially, this means that attempting to provision an encrypted volume after running a helm upgrade will fail, as the contents of px-vol-encryption will not match what’s been set as the cluster key in portworx (done via post-install job).

This PR fixes this by moving the secret creation to a post install job , where it is created before the job to set the cluster wide key has been run. 

The result is a consistent px-vol-encryption secret.


harsh-px:
@sean-mcgimpsey can you update the description with the issue and how the change fixes it?

Overall it looks good. I will also give this PR a run in our internal automation (This isn't hooked into travis yet).

sean-mcgimpsey:
Thanks. I’ve upgraded the description , hope that helps.

PR #133, opened 2020-07-07
Added option to require px/enabled label on hosts.

sean-mcgimpsey:
Introduced a new value requirePxEnabledTag. 

When set to true, portworx workloads will only install across nodes when px/enabled: true label is set on node. 

This feature is useful in eks/autoscaling environments where you want to stop portworx from automatically rolling out to newly provisioned nodes (where the node count is higher than 3). Without this behaviour, portworx DS will always try and install across all available nodes even though the  cluster limit is dependant on license restrictions.


sean-mcgimpsey:
Any update on this one? If possible  I'd like to get this merged.

PR #136, opened 2020-07-10
Added clusterDomain and security context

tusharraut1994:


charts/px-backup/values.yaml
```diff
@@ -38,7 +45,7 @@ images:
 
   pxcentralApiServerImage:
     imageName: pxcentral-onprem-api
-    tag: 1.0.4-dev
+    tag: 1.0.4
```
disrani-px, line 48: Use 1.0.4-rc1

tusharraut1994: Changed.

PR #140, opened 2020-07-18
Added upgrade support for onprem operator based install to helm chart

tusharraut1994:

- Changed etcd object names to support upgrade path.
- Added flag `operatorToChartUpgrade` to set previous install default values.


charts/px-backup/templates/px-backup/pxcentral-etcd.yaml
```diff
@@ -6,10 +6,10 @@
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: px-backup-etcd-scripts
+  name: pxc-backup-etcd-scripts
```
harsh-px, line 9: I would prefer to keep `pxc` reserved for all the `pxc` (The upcoming PX CLI) project.

Since both of them will be used in the same cluster, best to have separate naming convention.

You can use `pxcentral` instead.

charts/px-backup/templates/px-central-ui/pxcentral-ui.yaml
```diff
@@ -245,7 +246,11 @@ metadata:
   namespace: {{ .Release.Namespace }}
 type: Opaque
 stringData:
-  DB_PASSWORD: "Password1"
+  {{- if eq $operatorToChartUpgrade true }}
+  DB_PASSWORD: singapore
+  {{- else }}
+  DB_PASSWORD: Password1
```
harsh-px, line 252: just for my knowledge, what is the background behind the password change? And why `singapore` specifically?

charts/px-backup/templates/px-central-ui/pxcentral-ui-configmap.yaml
```diff
@@ -14,7 +15,11 @@ data:
   DB_PORT: "3306"
   DB_USERNAME: root
   DB_DATABASE: pxcentral
+  {{- if eq $operatorToChartUpgrade true }}
+  DB_PASSWORD: singapore
+  {{- else }}
   DB_PASSWORD: Password1
```
disrani-px, line 21: We should store this in a secret and allow users to set this

disrani-px: Looks like this is already in the secret for mysql, the backend component should read it from there

charts/px-backup/templates/px-central-ui/pxcentral-ui.yaml
tusharraut1994: This password is used by UI backend MySQL DB, Binu has hardcoded it in the previous release. so to support operator to chart upgrade we need to make it configurable and add/paas old password value.

charts/px-backup/templates/px-backup/pxcentral-etcd.yaml
tusharraut1994: Yeah, we changed all components prefix to pxcentral but for etcd, we had an issue in migrating existing ETCD to chart one where endpoints are changed so reverted change for etcd and added `pxc` as prefix to support operator install to chart upgrade support.

charts/px-backup/templates/px-central-ui/pxcentral-ui-configmap.yaml
tusharraut1994: Yes, In MYSQL statefulset I am reading password of a secret but for UI backend Binu needs it in configmap. so once Binu adds to change for reading the password from secret I can remove password from configmap.

PR #141, opened 2020-07-18
Added volume snapshot from pvc restore specs

tusharraut1994:

Added specs to create volume snapshot from existing pvc's and create new pvc's from volume snapshot.


pxcentral-operator-to-chart-upgrade/create_new_pvc_from_snapshot.yaml
```diff
@@ -0,0 +1,149 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: pxc-etcd-data-pxc-backup-etcd-0
+  namespace: portworx
+  labels:
+    app.kubernetes.io/component: px-backup-etcd
```
disrani-px, line 7: These aren't created by helm, so don't need these labels.

pxcentral-operator-to-chart-upgrade/README.md
```diff
@@ -0,0 +1,31 @@
+## Steps to upgrade operator based onprem central deployment to helm chart.
+
+1. Create volumesnapshot from existing onprem central pvc's in the same namespace.
+```console
+$ kubectl apply -f create_volume_snapshot_from_pvc.yaml
+```
+
+2. Verfiy volume snapshot for all pvc's are available in the same namespace.
+```console
+$ kubectl get volumesnapshot --namespace portworx
```
disrani-px, line 10: Use storkctl here, it'll show the status of the snapshot

pxcentral-operator-to-chart-upgrade/create_new_pvc_from_snapshot.yaml
tusharraut1994: yes, the reason for adding all labels to all pvc's because in helm chart we have labels but in operator install it was not there and when we try to install helm chart it gives errors because of the label does not exists and it tries to update pvc's but as its immutable object it fails to update PVC's

pxcentral-operator-to-chart-upgrade/README.md
tusharraut1994: Changed.

pxcentral-operator-to-chart-upgrade/create_new_pvc_from_snapshot.yaml
disrani-px: Some of these labels depend on the release name and the version of the chart, so there will be a mismatch, right?

tusharraut1994: yes, here we have some dependency like while helm chart installs we need to force the user to use release name as `px-backup` only and as this upgrade is only required for chart 1.0.0 release so there will be no version mismatch in labels. Because creating pvc's from a snapshot is the step needs to do before installing chart.

following things needs to update based on current px-central deployment:

create_new_pvc_from_snapshot.yaml
labels:
```
app.kubernetes.io/name: px-backup
app.kubernetes.io/version: 1.0.0
```
Annotations:
```
meta.helm.sh/release-name: px-backup
meta.helm.sh/release-namespace: portworx 
stork.libopenstorage.org/snapshot-source-namespace: portworx
```

create_volume_snapshot_from_pvc.yaml
`namespace: portworx`

harsh-px:
@tusharraut1994 is this still needed?

PR #142, opened 2020-07-20
Read username and password from values.yaml

tusharraut1994:

- read pxcentral database password from values.yaml
- updated image tags to use rc1 
- added pxcentral database username and keycloak username into values.yaml for secret templates.


charts/px-backup/templates/px-central-ui/pxcentral-ui-configmap.yaml
```diff
@@ -1,6 +1,5 @@
 {{/* Setting defaults if they are omitted. */}}
```
sharma-tapas, line 1: do we need this configmap if we are reading everything from the secrets?

tusharraut1994: Removed

PR #144, opened 2020-07-21
Added pod security context for postgresql statefulset to support upgrade

tusharraut1994:

We need to pod security context for PostgreSQL pod to support upgrade from operator based install to helm chart


charts/px-backup/templates/px-central-oidc/pxcentral-keycloak.yaml
```diff
@@ -267,12 +267,9 @@ spec:
           image: {{ printf "%s/%s/%s:%s" .Values.images.keycloakBackendImage.registry .Values.images.keycloakBackendImage.repo .Values.images.keycloakBackendImage.imageName .Values.images.keycloakBackendImage.tag }}
           imagePullPolicy: {{ .Values.images.pullPolicy }}
           {{ if eq $operatorToChartUpgrade true }}
-          {{- if $isOpenshiftCluster}}
```
sharma-tapas, line 270: what about fresh install on an openshift cluster. That will fail, as you wont add the security context?

tusharraut1994: For fresh install using helm chart we don't need podSecurity context, if `operatorToChartUpgrade` is set to `true` then only we need this.

PR #167, opened 2020-08-24
Updating monitoring templates and fixes for helm 3 (infix

### LINKED ISSUE #158: Portworx monitoring charts don't work with new releases of Kubernetes
**Is this a BUG REPORT or FEATURE REQUEST?**: BUG

**What happened**: Portworx Prometheus fails to start when deployment Portworx on Kubernetes 1.18.x

**What you expected to happen**: Expected Prometheus to run when deploying Portworx.

**How to reproduce it (as minimally and precisely as possible)**: Make sure you have "monitoring: true" in your values.yaml file and deploy Portworx on Kubernetes 1.18.x.

**Environment**:
- Container Orchestrator and version: N/A
- Cloud provider or hardware configuration: N/A
- OS (e.g. from /etc/os-release): N/A
- Kernel (e.g. `uname -a`): N/A
- Install tools: N/A
- Others: N/A

adasilva70:
This PR fixes 2 issues:
1. Update monitoring template to match latest generated by Portworx spec generator, more specifically updated Prometheus version to run on latest k8s releases and updated latest Grafana dashboards
2. Changed template to use `prefix` AND instead of `infix` so it can deployed with helm version 3 (this is because of GO 1.14 templates)

Closes #158

### DIFF
--- a/charts/portworx/templates/portworx-controller.yaml
+++ b/charts/portworx/templates/portworx-controller.yaml
@@ -1,5 +1,5 @@
 {{- $AKSorEKSInstall := or .Values.AKSInstall  .Values.EKSInstall  }}
-{{- if or ((.Values.openshiftInstall) and (eq .Values.openshiftInstall true)) ($AKSorEKSInstall) ((.Capabilities.KubeVersion.GitVersion | regexMatch "gke")) }}
+{{- if or (and (.Values.openshiftInstall) (eq .Values.openshiftInstall true)) ($AKSorEKSInstall) ((.Capabilities.KubeVersion.GitVersion | regexMatch "gke")) }}
 apiVersion: v1
 kind: ServiceAccount
 metadata:
--- a/charts/portworx/templates/portworx-monitor.yaml
+++ b/charts/portworx/templates/portworx-monitor.yaml
@@ -1,4 +1,4 @@
-{{- if (.Values.monitoring) and (eq .Values.monitoring true) -}}
+{{- if (and (.Values.monitoring) (eq .Values.monitoring true)) -}}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
@@ -37,6 +37,8 @@ rules:
       - prometheuses/finalizers
       - servicemonitors
       - prometheusrules
+      - podmonitors
+      - thanosrulers
     verbs: ["*"]
   - apiGroups:
       - apps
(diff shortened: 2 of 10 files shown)


charts/portworx/templates/portworx-controller.yaml
```diff
@@ -1,5 +1,5 @@
 {{- $AKSorEKSInstall := or .Values.AKSInstall  .Values.EKSInstall  }}
-{{- if or ((.Values.openshiftInstall) and (eq .Values.openshiftInstall true)) ($AKSorEKSInstall) ((.Capabilities.KubeVersion.GitVersion | regexMatch "gke")) }}
+{{- if or (and (.Values.openshiftInstall) (eq .Values.openshiftInstall true)) ($AKSorEKSInstall) ((.Capabilities.KubeVersion.GitVersion | regexMatch "gke")) }}
```
pault84, line 2: I don't know if this works as expected or not.

charts/portworx/templates/portworx-monitor.yaml
```diff
@@ -466,17 +459,17 @@ metadata:
...
       app: grafana
-  replicas: 1
   template:
     metadata:
       labels:
         app: grafana
     spec:
       containers:
-        - image: grafana/grafana:5.0.4
+        - image: grafana/grafana:5.3.3
```
pault84, line 472: we should bump this further. Px-Central installs 7.1.3 i believe @sharma-tapas ?
Let's give this a test with 7.1.3 and re-update the chart if it works. i know i have seen some security issues  with this version

charts/portworx/templates/portworx-controller.yaml
adasilva70: @pault84 those changes are needed for helm 3.3. I tested with helm version 3.3, 3.2.4 and 2.16.10

charts/portworx/templates/portworx-monitor.yaml
sharma-tapas: On-prem-ui is using the old version see here: [https://github.com/portworx/pxcentral-onprem-ui/blob/cc6ad56f2ca340618ee2eb428d8dbe19e5b10d00/dockerGroup/deployments/monitoring/grafana-deployment.yaml#L6064](https://github.com/portworx/pxcentral-onprem-ui/blob/cc6ad56f2ca340618ee2eb428d8dbe19e5b10d00/dockerGroup/deployments/monitoring/grafana-deployment.yaml#L6064 )
We should update both to use the latest 7.1.3, let me talk to Binu in case of PX-Backup monitoring I will confirm with @tusharraut1994 and update.

adasilva70: I did a quick test with grafana 7.1.3 and all dashboards look good.

adasilva70: @pault84 I just updated image to 7.1.3. Can this PR be merged now?

PR #181, opened 2020-09-18
Add missing pvc expander

duncangrist:

This PR adds the missing "persistentvolume-expander" controller to portworx-pvc-controller deployment definition.

This brings the Helm chart inline with yaml generated by PX Central.

Without this in place, Portworx volumes are unable to be expanded from the PVC request definition; you get the following error when describing the PVC if you try:

Error expanding volume "my-namespace-name/my-pvc-name" of plugin "kubernetes.io/portworx-volume": Get http://x.x.x.x:9001/v1/osd-volumes/versions: dial tcp x.x.x.x:9001: i/o timeout

This change fixes that.


duncangrist:
Closed this as it had more changes in it than I expected - I'll cherrypick out the specific change and re-raise.

PR #196, opened 2020-11-19
reading namespace name dynamically for the environment variables for restores of px-central namespace to succeed.

diptiranjanpx:
Signed-off-by: Diptiranjan Sahoo dipti@portworx.com

1. namespace names in env variables have been changed from Release.Namespace to metadata.namespace to avoid the fixed value of namespace name.
2. etcd endpoints to remove the namespace dependency.

Closes #
1. Fixes the issue where we restore the namespace having the central resources still pointing to the source namespace name instead of referring to the new namespace name in the destination cluster.


sharma-tapas (commented):
You need to add the helm package to the PR.

diptiranjanpx:
> You need to add the helm package to the PR.

Done.

tusharraut1994 (approved):
LGTM

@diptiranjanpx please make sure install and upgrade work fine.

PR #197, opened 2020-11-25
Updated pxcentral api server role to allow access to all resources fr…

tusharraut1994:
…om same namespace


diptiranjanpx (approved):
LGTM

sharma-tapas (commented):
why did we remove the file from staging, how are we testing this PR?

PR #200, opened 2020-12-01
OC-526 Added revision history to cleanup old replicaset on deployment…

tusharraut1994:
… update


diptiranjanpx (approved):
The only limitation we will get into is these deployments could not be rolled back.

sharma-tapas (approved):
lgtm

PR #208, opened 2020-12-08
Setting proper rediction uri for /auth.

diptiranjanpx:
Signed off: Diptiranjan

Fixing the redirection URI for /auth after session timeout.


tusharraut1994 (approved):
LGTM

PR #219, opened 2021-01-21
helm chart upgraded to 1.2.2

diptiranjanpx:
Signed off - Diptiranjan
 
px-backup - 1.2.2
px-monitor - 1.2.1
px-license-server - 1.2.1


charts/px-backup/README.md
```diff
@@ -37,12 +37,27 @@ Helm 2:
 $ helm install --name px-backup portworx/px-backup --namespace px-backup
 ```
 
+## Enabling/Disabling px-backup
```
bhavana-ar, line 40: PX-backup can now be disabled while installing px-central using helm chart version 1.2.2 onwards . It will be enabled by default. To disable PX-backup add the following  to your helm install command 
--set pxbackup.enabled=false
To enable px-backup after installing px-central, Follow the upgrade steps mentioned below, but either change pxbackup.enabled parameter in the values.yaml to "true"  or pass --set pxbackup.enabled=true to the helm upgrade command in Step 4

bhavana-ar (approved):

charts/px-backup/README.md
diptiranjanpx: Done.

PR #221, opened 2021-01-27
 Removing the dynamic labels in spec.selector.Matchlabels.

diptiranjanpx:
Signed off: Diptiranjan

Upgrade was facing the issue as the MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable


charts/px-monitor/templates/prometheus/prometheus.yaml
```diff
@@ -181,7 +181,6 @@ spec:
     matchLabels:
       k8s-app: pxcentral-prometheus-operator
       app.kubernetes.io/component: pxcentral-prometheus
-{{- include "px-monitor.labels" . | nindent 6 }}
```
sharma-tapas, line 184: what will happen to the `run:` label that identifies this?

diptiranjanpx: Did not see any component using 
run: px-monitor or run: px-license-server. 
Moreover, the selector.matchlabels are for the deployment to find its own resource like pods.

sharma-tapas (approved):

PR #222, opened 2021-02-09
Added all templates into single chart

diptiranjanpx:
Signed off:  Diptiranjan

This is the support for merging all three charts px-backup, px-monitor, and px-license-server into one chart.
Updated labels into all templates
Removing the pre and post scripts for monitor and ls.


sharma-tapas (approved):
too big of a review, approving for the sprint timeline will be still reviewing the code.

disrani-px (changes requested):
Why are all the charts being merged into one? This will create issues during upgrade and also we wanted to keep the px-backup chart separate

diptiranjanpx:
> Why are all the charts being merged into one? This will create issues during upgrade and also we wanted to keep the px-backup chart separate

The single chart approach was the one that was decided among the different other implementations.

PR #223, opened 2021-02-09
volumes in px-central should override the default values with user provided values.

diptiranjanpx:
Signed Off: Diptiranjan


sharma-tapas (approved):
lgtm

PR #249, opened 2021-04-20
[OC-284]: Supporting clusterUUID in grafana dashboards.

diptiranjanpx:
Signed-off By: Diptiranjan

The new dashboard screenshots are attached in the issue OC-284.


sharma-tapas (commented):
How does this work for old installs of Portworx that do not have ClusterUUID in the Prometheus metrics?

diptiranjanpx:
> How does this work for old installs of Portworx that do not have ClusterUUID in the Prometheus metrics?

Essentially the fix is the following.
1. There is a new variable clusterUUID (default value is all(.*) and if passed by the URL, it will take this value).
2. So if the URL has only clusterUUID, that clusterUUID and the corresponding cluster gets selected.
3. if the URL has the only cluster name, then directly the cluster gets selected.

Have put the details in OC-284. So in this way, if the cluster name does not exist even if UUID is correct still it will not show any data for the cluster belonging to the clusterUUID. So UI has to take the decision of sending only cluster name or clusterUUID based on the px version.

sharma-tapas (approved):

PR #258, opened 2021-05-18
[OC-744]: Migration script to support air-gapped clusters.

diptiranjanpx:
Signed-off by: Diptiranjan

Tests:
1. Normal upgrade which will require the values.yaml file.
2. Mongo trial migration with custom registry and repo.
3. Rollback after migration failure with custom registry and repo.

Test results have been published in OC-744.


single_chart_migration/migration.sh
```diff
@@ -47,17 +47,35 @@ cortex_nginx_deployment="pxcentral-cortex-nginx"
 ls_configmap="pxcentral-ls-configmap"
 post_install_job="pxcentral-post-install-hook"
 
+job_registry="docker.io"
+job_repo="portworx"
+job_image="pxcentral-onprem-post-setup"
+job_imagetag="1.3.0-dev"
```
sharma-tapas, line 53: We need to remember to make the image tag to 1.3.0

sharma-tapas (approved):

single_chart_migration/migration.sh
diptiranjanpx: yes, maybe I will create another PR and keep it ready for changing these constants.

PR #259, opened 2021-05-21
PWX-20143: updated the rbac config for the csi cloud drives support

trierra:

Closes #


charts/portworx/templates/portworx-rbac-config.yaml
```diff
@@ -9,37 +9,44 @@ metadata:
 
 kind: ClusterRole
 apiVersion: {{ template "rbac.apiVersion" . }}
-metadata:
-   name: node-get-put-list-role
```
harsh-px, line 13: why is name removed? this won't work.

harsh-px (changes requested):

charts/portworx/templates/portworx-rbac-config.yaml
harsh-px: I pushed a commit to fix this

harsh-px (approved):

PR #269, opened 2021-06-30
Added mongoCacheSize helm parameter to configure the cache size for mongo server

siva-portworx:
Added mongoCacheSize helm parameter to configure the cache size for mongo server. Default value is 4G.

Closes # PR:1646

Testing:
```
helm install central px-central --namespace px-backup --create-namespace --set persistentStorage.enabled=true,persistentStorage.storageClassName=mysql-sc,pxbackup.enabled=true,pxbackup.datastore=mongodb,persistentStorage.mongoCacheSize=1
```
```
[root@central-big-jackal-2 helm]# kubectl get sts pxc-backup-mongodb -n px-backup -o yaml | grep wiredTigerCacheSizeGB
          value: --wiredTigerCacheSizeGB=1
[root@central-big-jackal-2 helm]#
```


prashanthpx:
Changes look good, shouldn't you also build the helm package as part of this PR so that one can use it from 1.3.0 branch or that would be done separately?

siva-portworx:
> Changes look good, shouldn't you also build the helm package as part of this PR so that one can use it from 1.3.0 branch or that would be done separately?

Pushed the pkg as well

diptiranjanpx (approved):

PR #277, opened 2021-07-08
Moving px-central under charts directory.

diptiranjanpx:
Signed-off-by: Diptiranjan

Brought px-central chart under charts directory.

**Tests** 
Have tested install and upgrade cases.


sharma-tapas (approved):
I think the tar.gz file is missing

diptiranjanpx:
> I think the tar.gz file is missing

following is the tar bundle.

https://github.com/portworx/helm/blob/a6311772c88ea4fb95c3e3133b836089d527e698/stable/px-central-2.0.0.tgz

PR #281, opened 2021-07-19
[OC-950]: cassandra pvc size to default 64GB and in upgrade script to use the older size.

diptiranjanpx:
Signed-off-by: Diptiranjan

For the px-monitor backend which uses Cassandra as storage, 8GB seems to be less.
So suggesting the default PVC size to be 64GB.
For migration script, not to hit any failure because of this increase in size, migration script has been modified to use the old PVC size.

**Test**

Both install and upgrades have been tested and been updated in OC-950.


sharma-tapas:
Let's resolve the conflicts too.

sharma-tapas (approved):
LGTM, but let's also resolve the conflicts before merging.

PR #283, opened 2021-07-22
[OC-965]: Invalid params check in migration.sh

diptiranjanpx:
Mongo cache size to be 4GB for trial migration.

Signed-off-by: Diptiranjan

**Test**:
1. Successful migration to 2.0.0.
2. Input params validation.
3. Mongo trial migration

Detailed tests have been updated in OC-965.


single_chart_migration/mongo.yaml
```diff
@@ -179,6 +179,8 @@ spec:
               value: "no"
             - name: MONGODB_ENABLE_DIRECTORY_PER_DB
               value: "no"
+            - name: MONGODB_EXTRA_FLAGS
```
sharma-tapas, line 182: This change will also affect the final upgrade right.
You want it just for trial correct?

sharma-tapas (approved):

single_chart_migration/mongo.yaml
diptiranjanpx: In the trial run, this mongo.yaml gets used and once the migration is complete, the mongo resources including PVCs get deleted. 

For actual migration (which uses the helm package), the mongo cache size is already available as a helm param and the it can be configured. The default value for that is 4GB.

PR #284, opened 2021-07-23
PB-1778: Increase initialDelaySeconds for mongo pods.

adityadani:


diptiranjanpx:
Do we need this change for mongo trial migration too, then we have to change https://github.com/portworx/helm/blob/2.0.0/single_chart_migration/mongo.yaml

diptiranjanpx (approved):
LGTM

PR #290, opened 2021-08-11
OC-993 Increasing the buffer size as the response header is huge

prabhu-px:
OC-993 Increasing the buffer size as the response header is huge
Closes # OC-993


diptiranjanpx (commented):
Need helm package and index.yaml changes too.

diptiranjanpx (approved):
LGTM

PR #315, opened 2021-11-02
Update stork and pvc-controller for kubVer 1.22 

dahuang-purestorage:
Signed-off-by: dahuang <dahuang@purestorage.com>

Closes #


charts/portworx/templates/_helpers.tpl
```diff
@@ -22,7 +22,12 @@ release: {{ .Release.Name | quote }}
 {{- end -}}
 
 {{- define "px.kubernetesVersion" -}}
-{{$version := .Capabilities.KubeVersion.GitVersion | regexFind "^v\\d+\\.\\d+\\.\\d+"}}{{$version}}
```
adityadani, line 25: This function is being used for determining pvc-controller as well. So we might want to add a new function just for storkKubernetesVersion

dahuang-purestorage: updated

charts/portworx/templates/_helpers.tpl
```diff
@@ -25,6 +25,15 @@ release: {{ .Release.Name | quote }}
 {{$version := .Capabilities.KubeVersion.GitVersion | regexFind "^v\\d+\\.\\d+\\.\\d+"}}{{$version}}
 {{- end -}}
 
+{{- define "px.storkKubernetesVersion" -}}
+{{- $version := .Capabilities.KubeVersion.GitVersion | regexFind "^v\\d+\\.\\d+\\.\\d+" -}}
+{{- if semverCompare ">=1.22" $version -}}
+"v1.21.0"
```
adityadani, line 31: nit: Lets use v1.21.4, same as what we use in px-installer / spec-gen

P.S.: The [PR](https://github.com/portworx/px-installer/pull/1121) title says it is using 1.21.0, but we ended up using 1.21.4. Will update the title.

adityadani (approved):

charts/portworx/templates/_helpers.tpl
dahuang-purestorage: good catch :) updated

charts/portworx/templates/_helpers.tpl
```diff
@@ -25,6 +25,15 @@ release: {{ .Release.Name | quote }}
 {{$version := .Capabilities.KubeVersion.GitVersion | regexFind "^v\\d+\\.\\d+\\.\\d+"}}{{$version}}
 {{- end -}}
 
+{{- define "px.storkKubernetesVersion" -}}
```
piyush-nimbalkar, line 28: nit: Can we call this `storkSchedulerVersion` instead?

piyush-nimbalkar (approved):

charts/portworx/templates/portworx-controller.yaml
```diff
@@ -104,7 +104,9 @@ spec:
       - command:
         - kube-controller-manager
         - --leader-elect=true
+        {{- if (semverCompare ">=1.22" .Capabilities.KubeVersion.GitVersion) }}
```
piyush-nimbalkar, line 107: We need to remove the address field for k8s >= 1.22 and keep it as is for lower versions.
```
{{- if (semverCompare "<1.22" .Capabilities.KubeVersion.GitVersion) }}
```

charts/portworx/templates/portworx-controller.yaml
```diff
@@ -114,8 +116,17 @@ spec:
           httpGet:
             host: 127.0.0.1
             path: /healthz
+            {{- if semverCompare ">=1.22" .Capabilities.KubeVersion.GitVersion }}
+            {{- if $AKSorEKSInstall }}
```
piyush-nimbalkar, line 120: This needs to be done only for AKS, and not for EKS.

charts/portworx/templates/portworx-controller.yaml
```diff
@@ -114,8 +116,17 @@ spec:
...
+            port: {{ add .Values.pxStartPort 30 }}
```
piyush-nimbalkar, line 121: I checked the helm chart and looks like we have hard-coded the start port everywhere to 9001. I would recommend not adding the `pxStartPort` parameter right now because it is only going to be used here in the pvc-controller, while the rest of the places are not going to change. Just to avoid confusion lets remove the parameter and hard-code the value to `9031` here too.

charts/portworx/templates/portworx-controller.yaml
dahuang-purestorage: updated

charts/portworx/templates/portworx-controller.yaml
dahuang-purestorage: updated :)

charts/portworx/templates/portworx-controller.yaml
```diff
@@ -104,7 +104,9 @@ spec:
       - command:
         - kube-controller-manager
         - --leader-elect=true
+        {{- if (semverCompare "<1.22" .Capabilities.KubeVersion.GitVersion) }}
```
piyush-nimbalkar, line 107: Doesn't this require to strip out only the version part like you did for stork scheduler version?
```
.Capabilities.KubeVersion.GitVersion | regexFind "^v\\d+\\.\\d+\\.\\d+
```

dahuang-purestorage: updated :)

PR #320, opened 2021-11-16
[OC-1082]: migration script to work even if px-backup is disabled and for external ls server.

diptiranjanpx:
signed-off-by: Diptiranjan

1. Migration script to work even if px-backup is not installed and with external LS server.
2. Changes to support 2.1.0 migration.

### DIFF
--- a/single_chart_migration/migration.sh
+++ b/single_chart_migration/migration.sh
@@ -144,7 +144,7 @@ set_enabled_modules() {
         pxmonitor_enabled=true
     fi
 
-    ls_match=`$kubectl_cmd --namespace $namespace get deployment pxcentral-license-server`
+    ls_match=`$kubectl_cmd --namespace $namespace get configmap $cmListLS`
     if [ $? -eq 0 ] ; then
         pxls_enabled=true
     fi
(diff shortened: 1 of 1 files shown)


single_chart_migration/migration.sh
```diff
@@ -232,21 +231,26 @@ do_rollback() {
...
+    if [ `helm list -n px-backup | grep "px-central-2.0.1" | wc -l` -eq 1 ]; then
+        current_px_central_version="2.0.1"
+    elif [ `helm list -n px-backup | grep "px-central-2.0.0" | wc -l` -eq 1 ]; then
         current_px_central_version="2.0.0"
     fi
 
     # job_imagetag
-    if [ $current_px_central_version == "2.0.1" ]; then
-        # TODO: Needs to modify to public tags in 2.0.1 final patch
+    if [ $current_px_central_version == "2.1.0" ]; then
```
sharma-tapas, line 242: instead of using else if can we use an array or list of the version that is supported and check for that. 
In long run this will keep adding to the complexity, a better solution would be to say we support all versions ahead of 2.0.0

single_chart_migration/migration.sh
```diff
@@ -496,8 +500,8 @@ if [ "$helmvaluesfile" != "" ]; then
 fi
 
 if [ "$px_central_version" != "" ]; then
-    if [ $px_central_version != "2.0.0" ] && [ $px_central_version != "2.0.1" ]; then
-        echo "upgrade-version can only be 2.0.1 or 2.0.0"
+    if [ $px_central_version != "2.1.0" ] && [ $px_central_version != "2.0.1" ] && [ $px_central_version != "2.0.0" ]; then
```
sharma-tapas, line 503: These supported versions should be in an array or a list.

PR #337, opened 2022-02-16
2.1.2 dev ui images

prabhu-px:

Closes #


charts/px-central/values.yaml
```diff
@@ -126,17 +126,17 @@ images:
     registry: docker.io
     repo: portworx
     imageName: pxcentral-onprem-ui-frontend
```
diptiranjanpx, line 128: should not this image name be different for dev images?

diptiranjanpx:
Also please change the versions.yaml with the right images.

diptiranjanpx (approved):

PR #339, opened 2022-02-22
PWX-22962 update rbac for sharedv4-svc

dahuang-purestorage:
Signed-off-by: dahuang <dahuang@purestorage.com>

Updates the RBAC permission from px-installer spec to fix the permission issue with sharedv4 service. 

PWX-22962

Testing note:
Confirmed the existing charts shows the error. Updated the permission allows me to deploy a sharedv4 svc app.


charts/portworx/templates/portworx-rbac-config.yaml
```diff
@@ -23,13 +23,22 @@ rules:
...
+    verbs: ["get", "list", "create", "delete", "update"]
   - apiGroups: [""]
     resources: ["configmaps"]
     verbs: ["get", "list", "update", "create"]
+  - apiGroups: [""]
+    resources: ["services"]
+    verbs: ["get", "list", "create", "update", "delete"]
+  - apiGroups: [""]
+    resources: ["endpoints"]
+    verbs: ["get", "list", "create", "update", "delete"]
```
pureneelesh, line 41: nit: keep the order of verbs the same everywhere

nit: specify services and endpoints in the same section like pvc and pv

pureneelesh (approved):

charts/portworx/templates/portworx-rbac-config.yaml
dahuang-purestorage: this is the copied spec from px-installer. I think we should keep the same consistency between specs even if it's not best practices.

PR #361, opened 2022-05-31
Fix missing fields in StorageCluster template

haibinxie:


haibinxie:
@piyush-nimbalkar thanks for reviewing, this PR should resolve the comments in the previous PR.

charts/portworx/templates/storage-cluster.yaml
```diff
@@ -84,13 +84,21 @@ spec:
     kvdbDevice: {{ $kvdbDevice }}
     {{- end }}
 
+  {{- if .AKSInstall }}
+  cloudStorage:
```
piyush-nimbalkar, line 88: One thing we need to consider when "$drives" are specified is that it could be a cloud storage spec. We need some logic in the helm chart to decide whether to populate the `spec.storage` section `spec.cloudStorage` section.

piyush-nimbalkar (approved):
These changes look good.
We just need to fix the drives section when cloud storage specs are specified.

charts/portworx/templates/storage-cluster.yaml
piyush-nimbalkar: Yup. We do something similar in the px-installer/spec-generator.
https://github.com/portworx/px-installer/blob/a021d9c6739490d43a1e40e406f96c50faac30ae/px-spec-websvc/operator.go#L85
https://github.com/portworx/px-installer/blob/master/px-spec-websvc/templates/template_storage_cluster.gtpl#L72

piyush-nimbalkar: Feel free to do the change in another PR if you want.

PR #364, opened 2022-06-08
Keycloak update

prabhu-px:
Keycloak update to 16
Closes #


versions.yaml
```diff
@@ -7,7 +7,7 @@ modules:
     pxcentralOnpremUiLhbackend: "docker.io/portworx/pxcentral-onprem-ui-lhbackend:2.2.0"
     pxcentralOnpremPostSetup: "docker.io/portworx/pxcentral-onprem-post-setup:2.2.0"
     postgresql: "docker.io/portworx/postgresql:11.13.0-debian-10-r19"
-    keycloak: "docker.io/portworx/keycloak:14.0.0-v2"
+    keycloak: "docker.io/jboss/keycloak:16.1.1"
```
diptiranjanpx, line 10: Don't we have to build keycloak for this version?

diptiranjanpx: Also the image should be hosted in portworx repo.

PR #377, opened 2022-06-29
Portworx chart changes to support operator

piyush-nimbalkar:
- Support portworx uninstallation using the operator
- Add flags to install operator and storagecluster separately
- Fix annotation in the hook's RBAC objects
- Migration approved annotation needs to be added only if migration is
      needed (DS is present)
- Remove default stork version from values.yaml. This will ensure that the
      operator chooses the right component versions by default

Closes #

### DIFF
--- a/charts/portworx/templates/hooks/pre-delete/delete-storagecluster.yaml
+++ b/charts/portworx/templates/hooks/pre-delete/delete-storagecluster.yaml
@@ -1,0 +1,41 @@
+{{- if .Values.deployCluster }}
+{{- $customRegistryURL := .Values.customRegistryURL | default "none" }}
+{{- $registrySecret := .Values.registrySecret | default "none" }}
+{{- $clusterName := .Values.clusterName }}
+
+apiVersion: batch/v1
+kind: Job
+metadata:
+  namespace: kube-system
+  name: px-hook-delete-storagecluster
+  labels:
+    heritage: {{.Release.Service | quote }}
+    release: {{.Release.Name | quote }}
+    chart: "{{.Chart.Name}}-{{.Chart.Version}}"
+    app.kubernetes.io/managed-by: {{.Release.Service | quote }}
+    app.kubernetes.io/instance: {{.Release.Name | quote }}
+  annotations:
+    "helm.sh/hook": pre-delete
+    "helm.sh/hook-weight": "10"
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
+spec:
+  backoffLimit: 0
+  template:
+    spec:
+      {{- if not (eq $registrySecret "none") }}
+      imagePullSecrets:
+        - name: {{ $registrySecret }}
+      {{- end }}
+      serviceAccountName: {{ template "px.hookServiceAccount" . }}
+      restartPolicy: Never
+      containers:
+        - name: delete-storagecluster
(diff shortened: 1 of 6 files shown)


charts/portworx/templates/hooks/pre-delete/delete-storagecluster.yaml
```diff
@@ -0,0 +1,41 @@
...
+      containers:
+        - name: delete-storagecluster
+          {{- if eq $customRegistryURL "none" }}
+          image: "bitnami/kubectl:{{ template "px.kubectlImageTag" . }}"
+          {{- else}}
+          image: "{{ $customRegistryURL }}/bitnami/kubectl:{{ template "px.kubectlImageTag" . }}"
+          {{- end}}
+          command: ['/bin/sh',
+                    '-c',
+                    'kubectl -n kube-system delete storagecluster {{ $clusterName }}']
```
haibinxie, line 40: I think the goal is to wait for stc to disappear before deleting operator. storageCluster should be deleted by helm, the problem is operator is deleted hence storageCluster cannot be deleted.

charts/portworx/templates/serviceaccount-hook.yaml
```diff
@@ -6,8 +6,8 @@ metadata:
   name: {{ template "px.hookServiceAccount" . }}
   namespace: kube-system
   annotations:
-    "helm.sh/hook-delete-policy": before-hook-creation
-    "helm.sh/hook": "pre-delete,post-delete,post-install"
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
+    "helm.sh/hook": "post-install,pre-upgrade,pre-delete"
```
haibinxie, line 10: I think this should be post-delete, to let all resources be deleted before deleting the hook service account. I think I copied it from daemonset which has both pre-delete and post-delete, looks like it was a bug in daemonset chart.

charts/portworx/templates/hooks/pre-delete/delete-storagecluster.yaml
piyush-nimbalkar: By default, the `kubectl delete` command waits for the deletion to complete unless you pass `--no-wait` flag. So, this ensures that the storagecluster is deleted before the chart deletion proceeds. In the chart deletion, it will just delete the operator.

charts/portworx/templates/serviceaccount-hook.yaml
piyush-nimbalkar: The `helm.sh/hook` annotation defines what hook this service account is used for. As we added a pre-delete hook for cluster deletion, we need to use pre-delete and not post-delete. We don't need a post-delete hook as of now.

charts/portworx/templates/hooks/pre-delete/delete-storagecluster.yaml
piyush-nimbalkar: > the problem is operator is deleted hence storageCluster cannot be deleted
That is why, we are deleting the StorageCluster in `pre-delete` and once that is completed the chart deletion will delete the operator.

haibinxie: nice, forgot the command will block.

haibinxie (approved):

pure-jliao (approved):

PR #391, opened 2022-07-11
Delete daemonset chart and use operator chart

haibinxie:
The new operator chart will support:
- Migrating from daemonse to operator
- Fresh install with operator


charts/portworx/templates/portworx-operator.yaml
```diff
@@ -0,0 +1,99 @@
+{{- if .Values.deployOperator }}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: portworx-operator
+  namespace: kube-system
```
dsupure, line 6: A very dumb question. Does it need to be in kube-system, can user install the operator and the resources in other namspaces?

charts/portworx/templates/storage-cluster.yaml
```diff
@@ -0,0 +1,238 @@
...
+    deviceSpecs:
+      {{- $driveNames := $drives | split ";" }}
+      {{- range $index, $name := $driveNames }}
+      - {{ $name }}
+      {{- end }}
+{{- else }}
+  {{- if .AKSInstall }}
+  cloudStorage:
+    deviceSpecs:
+      - type=Premium_LRS,size=150
```
dsupure, line 98: Should we let user to customize the drive size?

charts/portworx/values.yaml
```diff
@@ -1,8 +1,12 @@
 # Please uncomment and specify values for these options as per your requirements.
 
+deployOperator: true                    # Deploy the Portworx operator
+deployCluster: true                     # Deploy the Portworx cluster
+
 deploymentType: oci                     # accepts "oci" or "docker"
 imageType: none                         #
 imageVersion: 2.9.1.4                   # Version of the PX Image.
```
dsupure, line 8: Should we change to 2.10.x?

charts/portworx/values.yaml
```diff
@@ -21,7 +25,7 @@ managementInterface: none             # Name of the interface <ethX>
 envVars: none                         # NOTE: This is a ";" seperated list of environment variables. For eg: MYENV1=myvalue1;MYENV2=myvalue2
 
 stork: true                           # Use Stork https://docs.portworx.com/scheduler/kubernetes/stork.html for hyperconvergence.
-storkVersion: 2.7.0
+storkVersion:                         # Version of Stork. For eg: 2.7.0
```
dsupure, line 28: We need a default value

charts/portworx/values.yaml
```diff
@@ -63,6 +72,7 @@ clusterToken:
   secretName: px-vol-encryption    # Name of kubernetes secret to be created. Requires clusterToken.create to be true.
   serviceAccountName: px-create-cluster-token # Service account name to use for post-install hook to create cluster token
 
-replicas: 3                         # Replica count for pvc-controller, stork, stork-scheduler.
-
 #requirePxEnabledTag: true               # if set to true, portworx will only install on nodes with px/enabled: true label. Not required in most scenarios.
+
+deleteStrategy:                           # Delete strategy for the portworx cluster
+  type:                                   # Valid values: Uninstall, UninstallAndWipe
```
dsupure, line 78: We need a default value here.

charts/portworx/templates/portworx-operator.yaml
haibinxie: technically it could be in other namespace, but since daemonset only uses kube-system, we use it as default

charts/portworx/templates/storage-cluster.yaml
haibinxie: It wont be customizable here, for daemonset to operator migration we need to keep consistent.

for other customers with new installs, if we get new requests we can add this support by other values in values.yaml

charts/portworx/values.yaml
haibinxie: Operator uses manifest to fetch the version, we don't need to provide the value by default.

charts/portworx/values.yaml
haibinxie: Operator will decide default deleteStrategy if it's empty, the chart can leave it empty.

charts/portworx/values.yaml
dsupure: If we don't this this, can we remove the field? Leaving it blank may confuse users since `values.yaml` is used to provide the default value.

charts/portworx/values.yaml
dsupure: Same question here. Can we delete it if it's taken care by the operator?

haibinxie: we still need it so users can change it.

charts/portworx/values.yaml
haibinxie: the behavior is different, when stork version is not provided, operator will get stork version according to px version based on version server or configmap (in airgapped environment). if stork version is provided, then it's pinned to the version, it won't change if px version changes. users can choose to pin it or not with this value.

charts/portworx/values.yaml
dsupure: If this is the case, we need to create different helm chart release version like what we do for the PSO. Customers won't know which version to specify. @piyush-nimbalkar : what's your suggestion?

haibinxie: I will bump up the version of both in another PR.

piyush-nimbalkar: @dsupure Looks good what @haibinxie is suggesting. We generally bump up the chart version with the default px version. 
About the version, we want to stick to 2.9.1.4 because that is the default version in our spec generator too - https://install.portworx.com/version. When we update the spec gen, we can update the helm chart too.

haibinxie:
@dahuang-purestorage I am going to merge this PR to master, any concerns?
@dsupure all of your comments have been answered or resolved, plz take another look.

PR #394, opened 2022-07-14
change version and remove unused values

haibinxie:

Closes #

### DIFF
--- a/charts/portworx/Chart.yaml
+++ b/charts/portworx/Chart.yaml
@@ -1,5 +1,5 @@
 name: portworx
-version: 2.11.0
+version: 2.9.1
 description: A Helm chart for installing Portworx on Kubernetes.
 tillerversion: ">=2.9.0"
 home: https://portworx.com/
--- a/charts/portworx/Chart.yaml
+++ b/charts/portworx/Chart.yaml
@@ -1,5 +1,5 @@
 name: portworx
-version: 2.9.1
+version: 2.11.0
 description: A Helm chart for installing Portworx on Kubernetes.
 tillerversion: ">=2.9.0"
 home: https://portworx.com/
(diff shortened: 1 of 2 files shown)


charts/portworx/values.yaml
```diff
@@ -3,9 +3,7 @@
 deployOperator: true                    # Deploy the Portworx operator
 deployCluster: true                     # Deploy the Portworx cluster
 
-deploymentType: oci                     # accepts "oci" or "docker"
-imageType: none                         #
-imageVersion: 2.9.1.4                   # Version of the PX Image.
+imageVersion: 2.11.0                    # Version of the PX Image.
```
piyush-nimbalkar, line 6: We still default to 2.9.1.4 from spec generator, let's stick to that. 
https://install.portworx.com/version

charts/portworx/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 name: portworx
-version: 2.9.1
+version: 2.11.0
```
piyush-nimbalkar, line 2: Let's update the version based on the px version. Check the other comment.

piyush-nimbalkar (approved):
Looks good after addressing the comments.

charts/portworx/Chart.yaml
haibinxie: Ok, I will keep it unchanged then, probably dont need the version suffix .4

piyush-nimbalkar:
@haibinxie I believe we will have to update the index as well with the new artifact.

piyush-nimbalkar:
Actually ignore my comment. Let's update the artifact when we merge the operator migration branch and bump up the chart version then.

dsupure:
@piyush-nimbalkar , @haibinxie : yes, this is what we should do.
If you see the index, for the px, it's still using 2.9.1. We should bum up the chart version and starts to make operator helm installation as the default. 
https://github.com/portworx/helm/blob/master/stable/index.yaml

PR #412, opened 2022-07-22
change internalKVDB to true by default

haibinxie:

Closes #

### DIFF
--- a/charts/portworx/values.yaml
+++ b/charts/portworx/values.yaml
@@ -66,7 +66,7 @@ deployOnMaster:  false                # For POC only
 csi: false                            # Enable CSI
 aut: false                            # Enable AutoPilot
 
-internalKVDB: false                   # internal KVDB
+internalKVDB: true                    # internal KVDB
 kvdbDevice: none                      # specify a separate device to store KVDB data, only used when internalKVDB is set to true
 
 etcd:                                 # DEPRECATED: Use kvdb.authSecretName for configuring secure etcd


pure-jliao (approved):

charts/portworx/values.yaml
```diff
@@ -66,7 +66,7 @@ deployOnMaster:  false                # For POC only
 csi: false                            # Enable CSI
 aut: false                            # Enable AutoPilot
 
-internalKVDB: false                   # internal KVDB
+internalKVDB: true                    # internal KVDB
```
piyush-nimbalkar, line 69: Let's change the default in the storagecluster template as well - https://github.com/portworx/helm/blob/operator-migration/charts/portworx/templates/storage-cluster.yaml#L4.

piyush-nimbalkar (changes requested):
Won't this break migrated customers who have etcdEndpoints set? We will default to internal kvdb for them and the etcd endpoints would be ignored. 
I guess we will have to update the logic in the storagecluster spec.

charts/portworx/values.yaml
haibinxie: values.yaml is just an example to use, it won't break existing customer. if we change the template then it would break customers who use a values.yaml that does not specify the flag. since it's already released to Mambu let's not change the template.

haibinxie:
> Won't this break migrated customers who have etcdEndpoints set? We will default to internal kvdb for them and the etcd endpoints would be ignored. I guess we will have to update the logic in the storagecluster spec.

the migrated customer should have no impact as they need to upgrade based on their original values.yaml through "helm get values".

piyush-nimbalkar:
> > Won't this break migrated customers who have etcdEndpoints set? We will default to internal kvdb for them and the etcd endpoints would be ignored. I guess we will have to update the logic in the storagecluster spec.
> 
> the migrated customer should have no impact as they need to upgrade based on their original values.yaml through "helm get values".

The values.yaml is not just an example file. It acts as the default value if no explicit value is specified through `--set` or `--values`.
Helm reference: https://helm.sh/docs/chart_template_guide/values_files/

I just tried it on my cluster and the internal kvdb overrides the existing external etcd endpoints.

haibinxie:
> > > Won't this break migrated customers who have etcdEndpoints set? We will default to internal kvdb for them and the etcd endpoints would be ignored. I guess we will have to update the logic in the storagecluster spec.
> > 
> > 
> > the migrated customer should have no impact as they need to upgrade based on their original values.yaml through "helm get values".
> 
> The values.yaml is not just an example file. It acts as the default value if no explicit value is specified through `--set` or `--values`. Helm reference: https://helm.sh/docs/chart_template_guide/values_files/
> 
> I just tried it on my cluster and the internal kvdb overrides the existing external etcd endpoints.

Yes, if nothing is given then it would use the default values. since the original values.yaml with etcdEndpoint is not default values.yaml, upgrading from non default values.yaml to default values.yaml is ok to break.

haibinxie:
After discussion with Piyush, this will break migration, hence closing it, we will keep internalKVDB default to false.

PR #480, opened 2023-05-04
Update Portworx to 2.13.3 and add custom namespace support

sharma-tapas:
This PR:
1. Update Portworx to install 2.13.3 and stork 23.4.0
2. Add Support for installing PX using operator in Custom Namespace


charts/portworx/templates/hooks/pre-delete/delete-storagecluster.yaml
```diff
@@ -6,7 +6,7 @@
 apiVersion: batch/v1
 kind: Job
 metadata:
-  namespace: kube-system
+  namespace: {{ template "px.getDeploymentNamespace" . }}-system
```
pragrawal-px, line 9: missed a `"-system"` here in the namespace value

charts/portworx/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 name: portworx
-version: 2.13.0
+version: 2.13.3
 description: A Helm chart for installing Portworx on Kubernetes.
 kubeVersion: ">=1.10.0"
 appVersion: "2.12.2"
```
pragrawal-px, line 5: for AWS charts, I have also been updating the appVersion values along with version.
Should that been done here as well

pragrawal-px (approved):

PR #502, opened 2023-08-14
pxcentral-backend pod init fix

t-nidhin:
During pxcentral-backend pod startup init container (init-mysql-db) calls keycloak service for default role mapping. This failed in case of a frontend url added to keycloak beforehand. 
This fix will allows to revert the frontend url in keyclaok and once the init script is completed, adds back the frontend url to keycloak.


charts/px-backup/templates/px-central-ui/pxcentral-ui.yaml
```diff
@@ -173,7 +174,10 @@ spec:
           command:
             - bash
             - -c
-            - cd /var/www/html && ./install.sh
+            - cd /var/www/html 
+            - python3 ./pre-init.py
```
diptiranjanpx, line 178: Can the whole things be put in install.sh?

PR #503, opened 2023-08-14
pxcentral-backend pod init fix

t-nidhin:
What this PR does / why we need it:
During pxcentral-backend pod startup init container (init-mysql-db) calls keycloak service for default role mapping. This failed in case of a frontend url added to keycloak beforehand.
This fix will allows to revert the frontend url in keyclaok and once the init script is completed, adds back the frontend url to keycloak.


diptiranjanpx (commented):
Modify in charts/px-central

t-nidhin:
> Modify in charts/px-central

done, moved changes to charts/px-central/pxcentral-ui.yaml

diptiranjanpx (approved):

PR #506, opened 2023-09-11
PB-4200 | Add Cache-Control headers for backup UI index page

kravichandran-px:

Closes # https://portworx.atlassian.net/browse/PB-4200

> Tested that index page is never cached. This is ensured by adding a Cache-Control header for index page in nginx.conf file.

> Adding query param in URL (part of this PR) to ensure that the cached page is not used after upgrade.

> Checked that none of the invocations of openUrl method (in which the change has been made) passes query param. So hard-coding `?v=2-5-2` in the method will not break anything.

> This is an interim solution. Will ensure in the next release that, 1) we cache-bust properly without the need for this query param and 2) never reload the entire app on navigating within the app (instead just use Angular SPA based routing).

https://github.com/pure-px/pxcentral-onprem-ui/assets/131302497/5c4cf41f-2657-4809-afad-1718d58f1741

### DIFF
--- a/charts/px-central/templates/px-lighthouse/px-central-ui/pxcentral-ui.yaml
+++ b/charts/px-central/templates/px-lighthouse/px-central-ui/pxcentral-ui.yaml
@@ -52,6 +52,7 @@ data:
       server_tokens off;
       location / {
         root   /usr/share/nginx/html;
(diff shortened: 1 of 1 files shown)


prabhu-08 (approved):

kravichandran-px:
Creating a new PR from 2.6.0-dev

PR #513, opened 2023-10-05
PWX-34239 Add support for GKE

nikolaypopov:
Add support for GKE


piyush-nimbalkar (approved):
Add it to the [Readme](https://github.com/portworx/helm/blob/master/charts/portworx/README.md) as well.

PR #520, opened 2023-11-09
PB-4860: Added checksum env for restarting keycloak pod when ca cert is set

px-kesavan:
**What this PR does / why we need it**: Added checksum env (HELM_CHECKSUM) for restarting keycloak pod when ca cert is set

Closes # https://portworx.atlassian.net/browse/PB-4860

Screenshot from 2023-11-09 17-30-14
Screenshot from 2023-11-09 17-37-09


kshithijiyer-px (approved):
LGTM

PR #555, opened 2024-04-03
PB-6455 :: Changing clusterRole to enable it access services too

vikasit12:
To find portworx namespace based on its service name 

Closes #PB-6455


charts/px-central/templates/px-backup/pxcentral-backup.yaml
```diff
@@ -28,7 +28,7 @@ rules:
     resources: ["jobs"]
     verbs: ["get", "list"]
   - apiGroups: [""]
-    resources: ["namespaces"]
+    resources: ["namespaces", "services"]
```
t-nidhin, line 31: referring to similar implementation in line 100 to 108, can we add this in a separate block even when they share similar permissions?
@prashanthpx Need your input here

PR #561, opened 2024-04-24
PB-6765: Added env QUERY_MONGO_DB to false for fresh installs

px-kesavan:
**What this PR does / why we need it**:  Added env QUERY_MONGO_DB to false for fresh installs

Closes # 


charts/px-central/templates/px-lighthouse/px-central-ui/pxcentral-lighthouse.yaml
```diff
@@ -148,6 +148,10 @@ spec:
           - name: SSL_CERT_DIR
             value: /tmp/certs
           {{- end }}
+          {{- if eq .Release.IsInstall true }}
+          - name: QUERY_MONGO_DB
```
kravichandran-px, line 152: Is this only for  middleware use case? If yes, should we rename this to be more specific to that use case?

charts/px-central/templates/px-lighthouse/px-central-ui/pxcentral-lighthouse.yaml
```diff
@@ -148,6 +148,10 @@ spec:
...
+            value: "false"
```
kravichandran-px, line 153: when will we set this to true?

kravichandran-px, line 153: lets create a task to remove this for all the further releases. Else I guess a customer who installed 2.7.0 fresh, but upgrades to 2.7.1 will end-up querying mongodb, which is not needed.

kravichandran-px (changes requested):

charts/px-central/templates/px-lighthouse/px-central-ui/pxcentral-lighthouse.yaml
px-kesavan: Yes, it only for middleware . This ENV is part of the middleware codebase . So i am reluctant in changing it

t-nidhin (approved):

PR #572, opened 2024-05-21
PB-6900: Added changes to bypass http proxy configuration

px-kesavan:
**What this PR does / why we need it**: Added the below helm params to bypass http proxy configuration
```
proxy:
  azureProxyEnabled: false
  httpProxy:
    noProxy: ""
```

proxy.azureProxyEnabled=true is set to use annotations on the specs to bypass http proxy in AKS
proxy.httpProxy.noProxy is provided with comma separated values to add no_proxy ENV in specs (applicable for any kubernetes discribution)

Note: the value of proxy.httpProxy.noProxy should be comma separated and those comma's should be escaped . Please refer below for a sample
--set proxy.httpProxy.noProxy="127.0.0.1\,localhost"

Closes # https://purestorage.atlassian.net/browse/PB-6900


charts/px-central/templates/px-backup/pxcentral-alertmanager.yaml
```diff
@@ -8,6 +9,11 @@ metadata:
   namespace: {{ .Release.Namespace }}
 spec:
   configSecret: px-backup-alertmanager-custom-config
+  {{- if eq $azureProxyEnabled true }}
+  podMetadata:
+    annotations:
+      kubernetes.azure.com/no-http-proxy-vars: "true"
```
prashanthpx, line 15: nit: can this annotation be taken from a variable? Good to do that if possible

charts/px-central/templates/px-monitor/grafana/grafana.yaml
```diff
@@ -192,11 +197,14 @@ spec:
         - name: pxcentral-grafana
           image: {{ printf "%s/%s/%s:%s" .Values.images.grafanaImage.registry .Values.images.grafanaImage.repo .Values.images.grafanaImage.imageName .Values.images.grafanaImage.tag }}
           imagePullPolicy: {{ .Values.images.pullPolicy }}
-          {{- if .Values.caCertsSecretName }}
+          {{- if or .Values.caCertsSecretName .Values.proxy.httpProxy.noProxy }}
```
prashanthpx, line 200: what is this change for?

prashanthpx (approved):

charts/px-central/templates/px-monitor/grafana/grafana.yaml
px-kesavan: This is required to add `env:` field or not. So we check for the existence of any one and add `env:`

charts/px-central/templates/px-backup/pxcentral-alertmanager.yaml
px-kesavan: This can be taken as an improvement, where we take any annotation from the User and apply it on all the specs

PR #595, opened 2024-07-04
PWX-37567 | Add missing fields for autopilot

hitesh-wani-px:
    Update autopilot component in helm to have all the fields present
Closes #https://purestorage.atlassian.net/browse/PWX-37567


charts/portworx/values.yaml
```diff
@@ -67,7 +67,26 @@ telemetry: false
 
 deployOnMaster:  false                # For POC only
 csi: false                            # Enable CSI
-aut: false                            # Enable AutoPilot
+
+
+autopilot: 
+  enabled: false                      # Enable AutoPilot
+  image: ""                           # Autopilot image
+  lockImage: false                    # Enables locking Autopilot to the given image.
```
nikita-bhatia, line 75: This field seems to be deprecated. Do we need to add this now?
@zoxpx please comment.

nikita-bhatia:
LGTM based on what we have on [Storage Cluster doc ](https://docs.portworx.com/portworx-enterprise/reference/crd/storage-cluster#autopilot-configuration) , but we also have Resources configuration in the autopilot spec.
Since the initial version of helm, we can add fields which are exposed in the doc, we can still go ahead.
@vinayakshnd @zoxpx can you folks comment if we need it as of now?

nikita-bhatia (approved):

PR #613, opened 2024-08-08
PB-7829: toleration for alert manager | px-backup | Alerting

t-nidhin:
This PR is for adding toleration for alert manager which is added to all in px-backup stack.


px-kesavan (approved):
LGTM

PR #621, opened 2024-08-13
PWX-38544 | Improvements for generic helm 

hitesh-wani-px:
**Improvements**
1. Update helm notes  which has incorrect links and readme file
2. Remove PX_SECRETS_NAMESPACE env variable 
3. Following PX components should be enabled by default in the values.yaml
    1. autopilot
    2. CSI
    3. internalKvdb
   
Closes #https://purestorage.atlassian.net/browse/PWX-38544


nikita-bhatia:
@hitesh-wani-px please re-check once if we really want to remove PX_SECRETS_NAMESPACE  if it was part of helm already, as I don't see it is deprecated.

charts/portworx/templates/storage-cluster.yaml
```diff
@@ -292,6 +292,9 @@ spec:
     {{- end }}
   {{- end }}
 
+  {{- $isLicenseSecretAdded := ne $licenseSecret "none" }}
+  {{- $shouldRenderEnv := or (ne $envVars "none") .Values.envs $AKSInstall $OKEInstall $isLicenseSecretAdded }}
```
diptiranjanpx, line 296: Is the following limited to AKS and OKE?

diptiranjanpx (approved):

charts/portworx/templates/storage-cluster.yaml
hitesh-wani-px: Yes some env var are that needs to be included are related to AKS and OKE only

hitesh-wani-px:
> PX_SECRETS_NAMESPACE



> @hitesh-wani-px please re-check once if we really want to remove PX_SECRETS_NAMESPACE if it was part of helm already, as I don't see it is deprecated.

The PX_SECRETS_NAMESPACE environment variable is currently being set by default, which should not be the case. It is not added as a default environment variable in the spec generation as well. If users require this variable, they can add it later using the env field. Removing this variable will not affect the installation process. To align with the spec generation, we should remove it as a default.

nikita-bhatia (approved):
LGTM

PR #635, opened 2024-09-04
PB-8057 | handle lost+found directory

ayush-px:
-> handle lost+found directory error by renaming the directory to lostfound


px-kesavan (approved):
LGTM

PR #704, opened 2025-01-19
PB-9130: Target px-backup version for install hook as well

kgarg-px:
While the generation of the report, we will also need to display the version of the px-backup being installed, to retrieve that information in case of installations, this information is vital
Closes #
PB-9130


mkoppal-px (approved):
LGTM

PR #706, opened 2025-01-21
PB-9422 : Moved random secret generation to hook requried for argoCD

px-kesavan:
**What this PR does / why we need it**: Moved random secret generation to hook requried for argoCD, 
Currently the password for prometheus and alertmanager is generated through helm function , Incase of argoCD, helm lookup function does not provide the result and it leads to regenerate the password during upgrade. 

Closes # https://purestorage.atlassian.net/browse/PB-9422


mkoppal-px (approved):
LGTM

PR #707, opened 2025-01-22
PB-9495: Added changes to change hook weight of pre install/upgrade hook pvc

px-kesavan:
**What this PR does / why we need it**: ArgoCD deploys the resources based on hook weights , in case of `volumeBindingMode: WaitForFirstConsumer` storageclasses , the pvc resource does not go into successful state in ArgoCD , unless the pod uses it. Inorder to fix it , the hook weight of hook pvc and hook job has been changed to hold the same hook weight

Closes # https://purestorage.atlassian.net/browse/PB-9495

Screenshot 2025-01-22 at 11 18 02 AM


ss-px (approved):
If this scenario is specific to only ArgoCD environments, can we use the `isArgoCD` param to set the weights?
Else, LGTM.

px-kesavan:
@ss-px , the changes done here doesnt affect the helm based deployment , so avoided usage of param to branch out the logic

PR #710, opened 2025-01-22
PB-3604: Remove default value for pullsecret (docregistry-secret)

px-kesavan:
**What this PR does / why we need it**: Remove default value for pullsecret (docregistry-secret) . This avoid warning in the events on production 

Closes # https://purestorage.atlassian.net/browse/PB-3604

Screenshot 2025-01-22 at 4 03 25 PM
Screenshot 2025-01-22 at 5 04 39 PM
Screenshot 2025-01-22 at 5 04 52 PM

Tested in MKE
Screenshot 2025-01-23 at 4 42 41 PM


charts/px-central/values.yaml
```diff
@@ -204,8 +204,7 @@ securityContext:
 tolerations: []
 
 images:
-  pullSecrets:
-    - docregistry-secret
+  pullSecrets: []
```
prashanthpx, line 207: When i was discussing with @diptiranjanpx on the original issue, he was saying empty string here was causing issue on MKE. Can we check once on MKE?

px-kesavan: @prashanthpx , It should be an empty array instead of empty string , will check it in MKE

dbinnal-px (approved):

mkoppal-px (approved):
LGTM

charts/px-central/values.yaml
px-kesavan: Tested in MKE, the deployment was successful , have attached in the PR description
@prashanthpx

PR #723, opened 2025-02-27
PB-7279: Support mtls for px-backup

px-kesavan:
**What this PR does / why we need it**: This PR has the following changes to support mtls through istio
1. Convert init container logic to main conatiners ,as there is a open issue in network connectivity between init container and services. Additionally, we will be maintaining a custom keycloak image for shipping nc binary into it
2. Added istio related CR configuration required for mtls
3. Handle upgrade by modifying the env, so the sts and deployments will be restarted when istio.enabled=true is set
4. Modified prometheus configurations , as there is a open issue in istio https://istio.io/latest/docs/ops/integrations/prometheus/
5. Modified nginx configuration for the frontend to comply with istio's requirement

Closes # https://purestorage.atlassian.net/browse/PB-9480


kgarg-px (approved):
LGTM!

ss-px (approved):

charts/px-central/templates/px-backup/service-mesh/pxcentral-istio.yaml
```diff
@@ -0,0 +1,72 @@
...
+{{- if not (eq $deployDedicatedMonitoringSystem true) }}
+apiVersion: security.istio.io/v1
+kind: PeerAuthentication
+metadata:
+  name: allow-external-prometheus-scrape
+  namespace: central
+spec:
+  selector:
+    matchLabels:
+      app: px-backup
```
kravichandran-px, line 67: can we do this for keycloak as well so that requests to kc from backend pod go through?

PR #725, opened 2025-03-05
PB-9480: Added changes to serviceMonitor to avoid setting tls secret path in case of byop

px-kesavan:
**What this PR does / why we need it**: Added changes to serviceMonitor to avoid setting tls secret path in case of byop

Closes #


kgarg-px (approved):
LGTM

ss-px (approved):

PR #737, opened 2025-04-11
PWX-43196 | PWX-43220 | Add imagePullSecret in operator deployment and Fix security context issue

hitesh-wani-px:
This PR adds support for specifying `imagePullSecrets` in the Portworx Operator deployment, which is necessary when pulling images from private registries.

It also fixes an issue where the Prometheus pod failed to start when `securityContext.runAsNonRoot` was set to true in `values.yaml`. Previously, the chart only allowed setting `runAsNonRoot`, but didn't provide a way to specify other necessary fields like `runAsUser`. This caused the Prometheus pod to crash due to insufficient security context configuration.

With this fix, users can now define a full `securityContext` block in `values.yaml`, including `runAsUser` and other relevant fields, ensuring that Prometheus starts successfully when running as a non-root user.

Closes # [PWX-43196](https://purestorage.atlassian.net/browse/PWX-43196) and [PWX-43220](https://purestorage.atlassian.net/browse/PWX-43220)

[PWX-43196]: https://purestorage.atlassian.net/browse/PWX-43196?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PWX-43220]: https://purestorage.atlassian.net/browse/PWX-43220?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/charts/portworx/values.yaml
+++ b/charts/portworx/values.yaml
@@ -138,8 +138,10 @@ monitoring:
     volumeMounts:                     # Additional VolumeMounts for the Prometheus StatefulSet
     # - mountPath: /test
     #   name: additional-volume
-    securityContext:                  # Enable prometheus container run as a non-root user.
-      runAsNonRoot: false
+    securityContext:                  # Enable prometheus container run as a non-root user. 
+    # Provide securityContext fields as defined by Kubernetes (https://kubernetes.io/docs/tasks/configure-pod-container/security-context/)
+      # runAsNonRoot: false
+      # runAsUser: 1000
   telemetry: true                    # Enable telemetry
   grafana: false                      # Enable grafana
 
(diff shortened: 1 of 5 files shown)


charts/portworx/values.yaml
```diff
@@ -138,8 +138,10 @@ monitoring:
     volumeMounts:                     # Additional VolumeMounts for the Prometheus StatefulSet
     # - mountPath: /test
     #   name: additional-volume
-    securityContext:                  # Enable prometheus container run as a non-root user.
-      runAsNonRoot: false
+    securityContext:                  # Enable prometheus container run as a non-root user.
```
nikita-bhatia, line 141: what will be the default value if user doesn't provide?

nikita-bhatia (approved):
LGTM

charts/portworx/values.yaml
nikita-bhatia: Discussed offline

PR #775, opened 2025-07-02
Add ability to set priorityClassName, annotations and imagePullPolicy

rvdh:
- [x] Signed off your work as per the DCO.
- [x] Add unit-tests

This PR adds three parameters to the Portworx helm chart:
- annotations: to set annotations on the StorageCluster object
- imagePullPolicy: to set the imagePullPolicy for PX pods 
- priorityClassName: to set PriorityClassName to be passed to Podspec of px pods for it to be scheduled accordingly

### DIFF
--- a/charts/portworx/README.md
+++ b/charts/portworx/README.md
@@ -155,6 +155,9 @@ The following tables lists the configurable parameters of the Portworx chart and
 | `updateStrategy.minReadySeconds` | Minimum number of seconds that a pod should be ready before the next batch of pods is updated during a rolling update | 1 |
 | `updateStrategy.disruption.allow` | This field is used to enable or disable smart and parallel upgrade. Smart upgrade is disabled by default, Enable it by setting to false. we can use the `maxUnavailable` field to control the maximum number of Portworx nodes that can be upgraded at a time | None |
 | `updateStrategy.autoUpdateComponents` | Specifies the update strategy for the component images. Valid values: None, Once, Always | None |
+| `annotations` | Map of annotations to set on the StorageCluster object itself. | {} |
+| `priorityClassName` | PriorityClassName to be passed to Podspec of px pods for it to be scheduled accordingly | "" |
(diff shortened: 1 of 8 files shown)


charts/portworx/README.md
```diff
@@ -155,6 +155,9 @@ The following tables lists the configurable parameters of the Portworx chart and
 | `updateStrategy.minReadySeconds` | Minimum number of seconds that a pod should be ready before the next batch of pods is updated during a rolling update | 1 |
 | `updateStrategy.disruption.allow` | This field is used to enable or disable smart and parallel upgrade. Smart upgrade is disabled by default, Enable it by setting to false. we can use the `maxUnavailable` field to control the maximum number of Portworx nodes that can be upgraded at a time | None |
 | `updateStrategy.autoUpdateComponents` | Specifies the update strategy for the component images. Valid values: None, Once, Always | None |
+| `annotations` | Map of annotations to set on the StorageCluster object itself. | {} |
```
hitesh-wani-px, line 158: We don't have support to add annotations directly to the stc. However, most annotations are already supported in portworx chart. Which additional annotations are we want to set on the STC ?

hitesh-wani-px:
Please share the list of test cases or scenarios covered during testing

charts/portworx/README.md
rvdh: Hi, we set the following annotations on the stc object:

    # https://github.com/libopenstorage/operator/issues/354
    # annotation needed to use tcr.paas.local as registry
    operator.libopenstorage.org/common-image-registries: "gcr.io,k8s.gcr.io"
    portworx.io/disable-storage-class: "true"

    argocd.argoproj.io/sync-wave: "1"
    # This resource doesn't exist yet on initial sync
    argocd.argoproj.io/sync-options: SkipDryRunOnMissingResource=true

Mainly the first annotation is critical for us, to be able to use our own internal registry.

rvdh:
> Please share the list of test cases or scenarios covered during testing

Test cases:
* TestAnnotations
* TestPriorityClassName
* TestImagePullPolicy

PR #778, opened 2025-07-14
[MTLS] - PB-11445, 11446 - Added changes to support mtls in pxb devploymet

px-kesavan:
**What this PR does / why we need it**: Added changes to support mtls in pxb deployment. Introduced two new parameters for instructing PXB to work with the corresponding service meshes

```
istio.enabled
linkerd.enabled
```

Closes # https://purestorage.atlassian.net/browse/PB-11445
https://purestorage.atlassian.net/browse/PB-11446 

Screenshot 2025-07-14 at 11 22 17 AM


charts/px-central/values.yaml
```diff
@@ -252,7 +258,7 @@ images:
     registry: pure-artifactory.dev.purestorage.com/px-docker-remote
     repo: portworx
     imageName: keycloak
-    tag: 26.2.4
+    tag: 26.2.4_v2
```
mkoppal-px, line 261: Why did we change the keycloak version?

mkoppal-px: Is it because we need all this?
```
COPY --from=nc-builder /bin/nc /usr/bin/nc
COPY --from=nc-builder /lib64/libssl.so.3 /lib64/
COPY --from=nc-builder /lib64/libcrypto.so.3 /lib64/
COPY --from=nc-builder /lib64/libpcap.so.1 /lib64/
COPY --from=nc-builder /lib64/libibverbs.so.1 /lib64/
COPY --from=nc-builder /lib64/libnl-route-3.so.200 /lib64/
COPY --from=nc-builder /lib64/libnl-3.so.200 /lib64/
```

charts/px-central/templates/px-backup/pxcentral-backup.yaml
```diff
@@ -158,7 +158,7 @@ spec:
   - protocol: TCP
     port: 10001
     targetPort: 10001
-    name: rest-api
+    name: http-rest-api
```
mkoppal-px, line 161: Any reason why this was changed?

charts/px-central/templates/px-backup/service-mesh/pxcentral-istio.yaml
```diff
@@ -0,0 +1,72 @@
...
+  - '*'
+  http:
+  - match:
+    - uri:
+        prefix: /
+    route:
+    - destination:
+        host: px-central-ui.{{ .Release.Namespace }}.svc.cluster.local
+        port:
+          number: 80
```
mkoppal-px, line 56: If we do this, then any request that is coming to this gateway will be sent to px-central-ui service?
Now what should a user do if they want to expose the px-backup service to make API calls directly. This command alone will not suffice right?
```
kubectl expose svc px-backup --type NodePort --port 10001 --name px-backup-exposed -n px-backup
```

charts/px-central/templates/px-lighthouse/px-central-ui/pxcentral-ui.yaml
```diff
@@ -63,7 +66,11 @@ data:
         proxy_cache_bypass  $http_upgrade;
         proxy_set_header Upgrade           $http_upgrade;
         proxy_set_header Connection        "upgrade";
+{{- if .Values.istio.enabled }}
+        proxy_set_header Host              "pxcentral-backend";
+{{- else }}
```
mkoppal-px, line 71: Why was this needed?

charts/px-central/values.yaml
mkoppal-px: But then the makefile and dockerfile should be in the pxcentral-backend repo right? Is this the right place to keep it?

charts/px-central/templates/px-lighthouse/px-central-ui/pxcentral-ui.yaml
px-kesavan: istio has specific requirements on the network calls to happen. so handling it specifically for istio , for normal installation i am not going with this flow , as it can bring in multiple hops in the network calls

charts/px-central/templates/px-backup/service-mesh/pxcentral-istio.yaml
px-kesavan: Hmm, in this case , we need to create a virtual service for the px-backup api as well , will take care in the documentation

charts/px-central/values.yaml
px-kesavan: The only difference is the nc binary , lets have it as a common for non mtls and mtls , will be easier to maintain and no need to pull extra image for aitgapped installation. 
Placed the dockerfile here so we wont miss it during any new releases , also keycloak is a third party image so having it in the pxcentral-backend will not bring any benefit from this place

px-kesavan: Have moved the dockerfile to the backend repo @mkoppal-px

mkoppal-px (approved):

PR #782, opened 2025-07-29
PWX-45412: Helm changes to support instance volumes

svijaykumar-px:
**What this PR does / why we need it**: This PR is the required helm changes for instance volumes. It sets the following annotations in the stc when insatnceVolumeK8sUpgrade is set to true in values.yaml
`portworx.io/disable-non-disruptive-upgrade: "false"
    portworx.io/move-volume-replica-before-upgrade: "true"`

Closes # PWX-45412

### DIFF
--- a/charts/portworx/templates/storage-cluster.yaml
+++ b/charts/portworx/templates/storage-cluster.yaml
@@ -80,11 +80,10 @@ metadata:
     {{- if (lookup "apps/v1" "DaemonSet" "kube-system" "portworx") }}
     portworx.io/migration-approved: "false"
     {{- end }}
-    {{- if eq .Values.nonDisruptivek8sUpgrade true }}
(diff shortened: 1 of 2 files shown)


charts/portworx/templates/storage-cluster.yaml
```diff
@@ -83,6 +83,10 @@ metadata:
     {{- if eq .Values.nonDisruptivek8sUpgrade true }}
     portworx.io/disable-non-disruptive-upgrade: "false"
     {{- end }}
+    {{- if eq .Values.instanceVolumeK8sUpgrade true }}
```
nikita-bhatia, line 86: @svijaykumar-px In case we set nonDisruptivek8sUpgrade and instanceVolumeK8sUpgrade both, the annotation `portworx.io/disable-non-disruptive-upgrade: "false"` will be added twice, can you correct that?

PR #812, opened 2025-09-26
PB-12587: Added fix to avoid null assignment to kubernetes spec fields

px-kesavan:
**What this PR does / why we need it**: Latest version of ArgoCD is not accepting the null values of a kuberenetes spec field. For example

env:
{{ if "some case"}}
 - name: foo
    value: bar
{{- end}}

if "some case" is false , env will be empty and argo is treating it as null and prompting with issues. This behavior is not seen in normal helm installation, as helm safely ignores. This PR fixes the places where there is a possibility of null values

Closes # https://purestorage.atlassian.net/browse/PB-12587

Screenshot 2025-09-26 at 7 18 12 PM


mkoppal-px (commented):
1. Isn't the same fix needed in other specs like px-backup deployment where we have if condition inside the env section of the px-backup deployment?
2. Can you also test the BYOP flow and ensure that it is not broken in proxy scenario?

px-kesavan:
> 1. Isn't the same fix needed in other specs like px-backup deployment where we have if condition inside the env section of the px-backup deployment?
> 
>     2. Can you also test the BYOP flow and ensure that it is not broken in proxy scenario?

1. Argo cd is complaining only on the custom resource
2. The changes done in this PR is not related to BYOP

PR #888, opened 2026-04-07
PB-15055: Add federated flag for post-install-job spec if exist

aks-px:
This change is added to have federated flag in post-install-hook job for deciding during creation of roles.Currently we wil not be creating app-admin role for federated mode
Closes #

Testing added as part of corresponding operator [PR.](https://github.com/pure-px/pxcentral-onprem-operator/pull/328)
Will remove the test commit before merging


mkoppal-px (approved):
LGTM. Revert the image tags

px-kesavan (approved):

PR #891, opened 2026-04-13
Central deps upgrade

kravichandran-px:
Upgrade Keycloak, Postgres and MySQL

Closes #

Tested:
Fresh install 3.0.0
Install 2.11 -> Upgrade to 3.0.0
Test 3.0.0 with EntraID
Install 2.11 with Ingress + SSL -> Upgrade to 3.0.0


charts/px-central/values.yaml
```diff
@@ -268,16 +268,16 @@ images:
     tag: 3.0.0-dev
     module: pxCentral
   keycloakBackendImage:
-    registry: pure-artifactory.dev.purestorage.com/px-docker-remote
```
px-kesavan, line 271: Can we revert the registry change.

px-kesavan (approved):
Apart from the minor comment ,LGTM

charts/px-central/values.yaml
kravichandran-px: Fixed

PR #927, opened 2026-06-11
PB-14077: Grant px-backup RBAC to read kdmp-config ConfigMap

wdhopte-px:
Add cluster-scoped configmaps 'get' permission to px-backup-cluster-role
so that px-backup can read the kdmp-config ConfigMap in kube-system
namespace. This is required for the KDMP PriorityClass feature where
the PriorityClassName for KDMP job pods is read from the
KDMP_JOB_PRIORITY_CLASS_NAME key in the kdmp-config ConfigMap.

Without this permission, delete and maintenance jobs created by
px-backup fail to read the PriorityClass configuration.


wdhopte-px:
Not required

PR #931, opened 2026-06-17
PB-16295: Make PX-Backup telemetry work on OpenShift

t-nidhin:
## Summary

  PX-Backup telemetry (the edge-envoy migration from PB-15800) doesn't work on OpenShift out of the box: no Pure1 cert is minted and no logs/metrics are uploaded. This PR makes telemetry fully functional on OpenShift while remaining a no-op on vanilla Kubernetes, using a least-privilege nonroot SCC.

### Root cause

  OpenShift's default restricted-v2 SCC runs pods as a random namespace UID with no fsGroup, ignoring the chart's runAsUser/fsGroup. That breaks two writes:

  1. edge-envoy /config (cert never minted) — edge-envoy self-renders its SDS/TLS cert into the image-baked /config owned by the pure user (uid 10001, owner-writable only). Random UID → render permission denied → envoy never serves :12001 → ccm-go connection refused → empty cert.
  2. collector /var/cores (no data uploaded) — the log-collector writes logs+metrics to the /var/cores emptyDir; without fsGroup it stays root:root and the non-root collector can't write → empty uploads.

  OpenShift-only regression from the telemetry-envoy → edge-envoy migration (old envoy mounted read-only ConfigMaps, never wrote /config).

  Fix (all gated on isOpenshift / inert on vanilla)

  1. Run edge-envoy as runAsUser: 10001 unconditionally so it owns /config.
  2. Apply the pod securityContext (fsGroup: 1000) on OpenShift so /var/cores is group-writable.
  3. isOpenshift-gated RoleBinding granting the two telemetry SAs the nonroot SCC.

  Why nonroot (not anyuid / nonroot-v2)

  - vs anyuid: telemetry needs only a fixed non-root uid (10001) + fsGroup, not root. nonroot grants exactly that and forbids root — least-privilege.
  - vs nonroot-v2: the *-v2 SCCs only exist on OCP 4.11+; binding to a missing ClusterRole on older clusters would silently fall back to restricted-v2 and re-break telemetry. The v2 delta (no priv-esc, drop-ALL-caps, seccomp) is PSA hardening this workload doesn't need (non-root, no privileged ports, no
  capabilities), so v1 nonroot is equivalent here and portable across all supported OCP versions.

  Testing done

  Platforms: OCP 4.16 (pwx-ocp-224-205), OCP 4.20 (pwx-ocp-0-215), vanilla k8s 1.35 (sys1) — each prod + staging.

  OpenShift, nonroot, end-to-end: SCC=nonroot; edge-envoy uid 10001, others 1000; /config renders cleanly, cert minted (~1.8 KB, Pure1-issued); /var/cores = drwxrwsr-x root 1000 writable; collected logs from 11 pods (724.6 KB) + metrics saved, cycle ~3s; all four data types (logs, metrics, db, reporting) confirmed in
  Pure1 fuse; no container runs as root.

  Vanilla, unaffected: helm template isOpenshift=false → 0 SCC bindings; telemetry healthy, logs+metrics in fuse (prod+staging).

  Render checks: isOpenshift=true → exactly one scc:nonroot binding, no anyuid; isOpenshift=false → no SCC objects, runAsUser: 10001 unchanged.

  Notes

  - No-op off OpenShift.
  - Pre-existing/out-of-scope: a fresh-install Mongo one-shot-init race (PB-13595) can leave db/reporting empty until the collector is bounced once after MongoDB is up; logs+metrics unaffected. Tracked separately.


lalat-das (approved):
lgtm

PR #949, opened 2026-07-20
PB-16568: harden px-central MongoDB replica set for HA under load

pallav-px:
## Problem

Under cluster load the px-central MongoDB replica set intermittently collapses to `ReplicaSetNoPrimary`, so px-backup API calls fail with `server selection timeout` and in-flight backup/restore operations fail (surfacing as flakes in `BasicBackupAndRestoreWithParallelBackupSchedule`). See **[PB-16568](https://purestorage.atlassian.net/browse/PB-16568)**.

**Root cause:** the MongoDB liveness probe execs `mongosh` (Node.js cold-start + connect + auth) with a 5s timeout and asserts replica-set role. Because the pods are **BestEffort** (`resources: {}`), under node CPU/IO/network contention the `mongosh` exec cannot finish within 5s; after 6 failures kubelet kills the container — and when the killed member is the **primary**, this forces an election and the `ReplicaSetNoPrimary` window that fails px-backup.

## Reproduced and fix-validated on a live cluster

| Metric | Old (`mongosh` exec liveness) | New (`tcpSocket`) |
|---|---|---|
| Probe latency under identical CPU starvation | **14,432 ms** → fails 5s budget → kill | **6 ms** → passes |
| Member killed within 95s of starvation? | killed in ~50s (RESTARTS 0→1) | **not killed** (RESTARTS=0) |
| Role during starvation | primary killed → election | **kept PRIMARY** |
| px-backup `server selection timeout` errors | reproduced (8×) | **0** |

## Changes (chart `px-central`)

1. **Liveness probe → `tcpSocket:27017`** — process-liveness only, so it cannot false-fail under starvation and can never restart a healthy primary. The `isWritablePrimary || secondary` role check moves to the **readiness** probe (which only removes the pod from the Service, never restarts it); readiness `timeoutSeconds` widened `5 → 15`.
2. **Container resources** (`values.mongodbResources`) — moves the pods off **BestEffort**. Memory limit (`6Gi`) exceeds the 4 GB WiredTiger cache (`persistentStorage.mongoCacheSize`) so it isn't OOMKilled at the cache ceiling. Stops the starvation and stops the scheduler co-locating heavy workload onto the mongo nodes.
3. **`terminationGracePeriodSeconds: 120`** (`values.mongodbTerminationGracePeriodSeconds`) — lets a SIGTERM'd mongod checkpoint and step down cleanly instead of being SIGKILLed on grace expiry (which caused WiredTiger crash recovery + oplog rollback).
4. **`podAntiAffinity` default `false → true`** (already values-gated in the chart) + a new **PodDisruptionBudget** (`minAvailable: 2`, toggle `values.mongodbPodDisruptionBudget`) — members never co-locate, and voluntary disruption can never drop below quorum.

All new behavior is values-tunable; set `mongodbPodDisruptionBudget: false` / `podAntiAffinity: false` for single-/two-node dev clusters.

## Testing

- `helm lint` and `helm template` pass; rendered output verified for all four changes.
- Live A/B on a px-backup cluster (chart `px-central-3.1.0`): the old exec probe killed the primary under CPU starvation and reproduced the exact px-backup `ReplicaSetNoPrimary` error; after switching liveness to `tcpSocket`, the primary survived identical starvation with 0 restarts and px-backup logged 0 errors. Cluster restored to as-found state afterward.

## Notes / follow-ups (not in this PR)

- The `pxc-backup-mongodb-scripts` ConfigMap (holding `setup.sh`) is **referenced but not created** by this chart — the slow-rejoin fast-path fix belongs wherever that script is generated (px-central operator).
- Optional defense-in-depth: raise px-backup's mongo `serverSelectionTimeout` and add Torpedo retry on transient `server selection timeout`.

## Acceptance criteria (PB-16568)

- ✅ A liveness failure cannot, by itself, kill the current primary — met structurally by (1).
- ✅ Members not restarted/evicted on transient load — (1) + (2) + (3).
- ✅ Replica set retains a primary under the load profile — (2) + (4).

[PB-16568]: https://purestorage.atlassian.net/browse/PB-16568?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/charts/px-central/templates/px-backup/pxcentral-mongodb.yaml
+++ b/charts/px-central/templates/px-backup/pxcentral-mongodb.yaml
@@ -114,6 +114,11 @@ spec:
         {{- end }}
       {{- end }}
       serviceAccountName: pxc-backup-mongodb
+      # PB-16568: give a SIGTERM'd mongod time to checkpoint and step down cleanly.
+      # With the default 30s, a starved primary is SIGKILLed on grace expiry and
+      # returns via WiredTiger crash recovery + oplog rollback (un-checkpointed
+      # writes discarded).
+      terminationGracePeriodSeconds: {{ .Values.mongodbTerminationGracePeriodSeconds | default 120 }}
       {{- if $isOpenshiftCluster}}
       {{- else }}
       securityContext:
@@ -185,13 +190,17 @@ spec:
             - containerPort: 27017
               name: mongodb
               protocol: TCP
+          # PB-16568: liveness must only prove the process is alive. A mongosh exec
+          # (Node.js cold-start + connect + auth) cannot finish within a 5s timeout
+          # under node starvation, and asserting replica-set role also fails during
+          # elections -- so the old exec probe would kill a healthy PRIMARY, collapsing
+          # the replica set to ReplicaSetNoPrimary. A cheap TCP check cannot false-fail
+          # under CPU pressure (measured 6ms vs 14-19s for the exec). Role/writability
+          # lives in the readiness probe below, which only removes the pod from the
+          # Service and never restarts the container.
           livenessProbe:
-            exec:
-              command:
-              - bash
-              - -ec
-              - |
-                mongosh --eval 'db.adminCommand("ping") && db.hello().isWritablePrimary || db.hello().secondary' | grep -q 'true'            
+            tcpSocket:
+              port: 27017
             failureThreshold: 6
             initialDelaySeconds: 600
             periodSeconds: 10
@@ -208,7 +217,13 @@ spec:
             initialDelaySeconds: 60
             periodSeconds: 10
             successThreshold: 1
-            timeoutSeconds: 5
+            # PB-16568: widened from 5s -- a starved secondary should be given room
+            # before it is pulled from the Service, to avoid needless endpoint churn.
+            timeoutSeconds: 15
+          {{- if .Values.mongodbResources }}
+          resources:
+{{ toYaml .Values.mongodbResources | indent 12 }}
+          {{- end }}
           volumeMounts:
             - mountPath: /data/db
               name: pxc-mongodb-data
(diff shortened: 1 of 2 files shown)


pallav-px:
## Live-cluster reproduction & fix validation (full results)

Run on a live px-backup cluster (4 nodes, `v1.34.0`) running the **exact unfixed chart** (`px-central-3.1.0`): all 3 mongo pods `BestEffort`, `mongosh` liveness probe (5s timeout, `×6`, asserts RS role), `terminationGracePeriodSeconds: 30`, no PDB.

### 1. The probe has no headroom (baseline, idle)
The exact liveness command, timed in-pod on an idle node:
```
run 1: 886 ms   run 2: 895 ms   run 3: 935 ms      (budget = 5000 ms)
```
~0.9s with mongod healthy and unloaded — already ~1/5 of the budget spent on `mongosh` startup + auth alone.

### 2. Liveness kill reproduced (root cause)
Pinned a CPU hog to the **primary's** node (BestEffort mongo → CFS-starved). Re-timed the same probe:
```
run 1: 18868 ms  [ok]
run 2: 16401 ms  [server is in quiesce mode and will shut down]   ← kubelet already killing it
run 3: 12683 ms  [exit 137]
```
Kubelet events on the primary:
```
Warning  Unhealthy  Liveness probe failed: command timed out ... "mongosh --eval '...'" timed out after 5s
Normal   Killing    Container mongodb failed liveness probe, will be restarted
```
Pod `RESTARTS 0 → 1` in-place — the exact signature from the diag bundles. Failover then elected a new primary.

### 3. px-backup `ReplicaSetNoPrimary` symptom reproduced
Forced a >30s no-primary window (froze both secondaries + stepped down the primary). px-backup's own reconcilers logged the ticket error verbatim (**8×** across workers):
```
failed to retrieve organizations: server selection error: server selection timeout,
current topology: { Type: ReplicaSetNoPrimary, Servers: [
  { Addr: pxc-backup-mongodb-0...:27017, Type: RSSecondary, Average RTT: 1120330 },
  { Addr: pxc-backup-mongodb-2...:27017, Type: RSSecondary, Average RTT: 630727 },
  { Addr: pxc-backup-mongodb-1...:27017, Type: RSSecondary, Average RTT: 1736972 } ] }
```

### 4. Fix validated — same hardware, same starvation, only the probe changed
Patched liveness → `tcpSocket:27017` (this PR's change #1) and re-ran the **identical** CPU-hog starvation on the current primary:

| | Old (`mongosh` exec) | New (`tcpSocket`) |
|---|---|---|
| Probe latency under starvation | **14,432 ms** → fails → kill | **6 ms** → passes |
| Member killed in 95s? | killed in ~50s | **RESTARTS = 0, no Killing event** |
| Role during starvation | primary killed → election | **kept PRIMARY (health=1)** |
| px-backup errors during window | `ReplicaSetNoPrimary` | **0** |

Under the fix, the starved member correctly went `ready=false` (readiness pulled it from the Service) but was **never restarted** and **retained its PRIMARY role**, so no election fired and px-backup stayed connected throughout.

### 5. Cleanup
Cluster restored to as-found state after validation: liveness reverted to the shipped `mongosh` exec, repro namespace deleted, replica set healthy 3/3, RESTARTS back to 0.

*Note: `resources` (#2), `terminationGracePeriodSeconds` (#3), and PDB/anti-affinity (#4) target the starvation, unclean-shutdown, and quorum-loss paths respectively; #1 above is the decisive isolated proof since it alone stops the primary kill.*

PR #950, opened 2026-07-20
PB-16568: widen px-central MongoDB probe timeouts to 300s (interim fix)

pallav-px:
## Problem

Under cluster load the px-central MongoDB replica set collapses to `ReplicaSetNoPrimary`, so px-backup API calls fail with `server selection timeout` and in-flight backup/restore ops fail (flaky `BasicBackupAndRestoreWithParallelBackupSchedule`). See **[PB-16568](https://purestorage.atlassian.net/browse/PB-16568)**.

**Root cause:** the MongoDB liveness probe execs `mongosh` (Node.js cold-start + connect + auth) with a **5s** timeout. On a starved node the exec can't finish in 5s → after 6 failures kubelet kills the container → if it's the **primary**, that forces an election and the `ReplicaSetNoPrimary` window that fails px-backup.

## Scope of this PR (interim fix for 3.1.0)

Deliberately minimal, per review feedback (keep the functional `mongosh` health check rather than switch to a shallow TCP check, and avoid broader changes so close to GA):

1. **Liveness probe** — keep the `mongosh` exec, `timeoutSeconds` **5 → 300**.
2. **Readiness probe** — keep the `mongosh` exec, `timeoutSeconds` **5 → 300**.
3. **`terminationGracePeriodSeconds: 120`** (`values.mongodbTerminationGracePeriodSeconds`) — lets a SIGTERM'd mongod checkpoint + step down cleanly.

The 300s timeout removes the starvation-induced **false** failure (measured probe latency: ~0.9s idle vs 14–19s under CPU starvation) while preserving "is mongod up / reachable / authenticating / answering" as the health signal.

**Dropped from earlier revisions** (per review — @mkoppal-px): the PodDisruptionBudget (only guards voluntary disruption, not the involuntary failure here) and the `podAntiAffinity` default flip (hard constraint risks `Pending` pods on upgrade). Also dropped the container `resources`/QoS change to keep this strictly a probe-timeout fix.

## Known limitations → follow-up

This is a pragmatic stopgap, not the final design. Tracked separately for post-3.1.0 analysis:
- The liveness probe still asserts replica-set role (`isWritablePrimary || secondary`), which can fail during a prolonged **election** independent of the timeout.
- A heavy `mongosh` exec on the liveness path stays fragile under *extreme* starvation (300s is headroom, not a guarantee).
- Resource/QoS hardening (mongo ships BestEffort) is deferred.
- **A node-starvation run will validate whether the 300s timeout is sufficient.**

## Testing

- `helm lint` + `helm template` clean; rendered output verified.
- Root cause reproduced live (liveness kill + `ReplicaSetNoPrimary`, then fix behavior) — see comment below. A starvation run against this exact 300s config is the next validation step.

[PB-16568]: https://purestorage.atlassian.net/browse/PB-16568?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


charts/px-central/values.yaml
```diff
@@ -25,7 +25,28 @@ persistentStorage:
 
 storkRequired: false
 nodeAffinityLabel: ""
-podAntiAffinity: false
```
mkoppal-px, line 28: Should we change the default here?

charts/px-central/templates/px-backup/pxcentral-mongodb.yaml
```diff
@@ -258,4 +273,26 @@ spec:
         storageClassName: {{ .Values.persistentStorage.storageClassName }}
         {{- end }}
   {{- end }}
+{{- if .Values.mongodbPodDisruptionBudget }}
+---
+# PB-16568: protect replica-set quorum. Without a PDB, a node drain / voluntary
+# disruption can take down 2 of 3 members at once, leaving no primary. minAvailable=2
+# keeps a voting majority (and therefore a primary) available at all times.
+apiVersion: policy/v1
```
mkoppal-px: Why do we need this? PDB helps only in voluntary disruption and does not help when a node goes down.

charts/px-central/templates/px-backup/pxcentral-mongodb.yaml
```diff
@@ -185,13 +190,17 @@ spec:
...
+          # Service and never restarts the container.
           livenessProbe:
-            exec:
-              command:
-              - bash
-              - -ec
-              - |
-                mongosh --eval 'db.adminCommand("ping") && db.hello().isWritablePrimary || db.hello().secondary' | grep -q 'true'            
+            tcpSocket:
+              port: 27017
```
px-kesavan: Why cant we just increase the timeout to 60s or more,  to address the livenessprobe issue ?

here we are checking only the port is open or not , which will be true in most of the time. We are losing the main functionality of livenessprobe here . where we fail to restart the pod for actual errors. 

Old liveness probe actually checks mongod is up, reachable, authenticating, answering commands.

charts/px-central/values.yaml
pallav-px: reverted. 
Making it the default flips it to a hard (`requiredDuringSchedulingIgnoredDuringExecution`) constraint for everyone, which can leave a mongo pod `Pending` on upgrade for any cluster with <3 schedulable nodes (or where members currently share a node under the old soft default). Kept the default `false` (the existing `preferred`/soft spread); production can opt in explicitly.

charts/px-central/templates/px-backup/pxcentral-mongodb.yaml
pallav-px: removed the PDB. 
It  guards voluntary disruptions (drains/evictions via the eviction API) and does nothing for an involuntary node-down. Since PB-16568's failure mode is involuntary (liveness kill / eviction / node pressure), the PDB adds no protection for the actual incident. Dropped it.

charts/px-central/templates/px-backup/pxcentral-mongodb.yaml
pallav-px: few reasons a bigger timeout alone doesn't fix this, and why the "deep" check isn't actually lost:

**1. The bigger problem isn't just the timeout — it's the role assertion.** The old probe requires `isWritablePrimary || secondary`. During any election a healthy member is transiently *neither* (STARTUP2/RECOVERING/candidate), so it fails the probe **regardless of the timeout value**. A 60s timeout wouldn't have prevented the kills — during an election storm the probe returns "not primary/secondary" quickly and still trips `failureThreshold`. A liveness probe should never assert replica-set role, because restarting a member mid-election makes the election worse.

**2. You can't size a timeout against node starvation.** Measured on a live cluster: the `mongosh` exec is ~0.9s idle but **14–19s** under a single CPU hog, and the ticket's real load (90 MiB/s throttle + mysqlslap co-located) is heavier — it could exceed 60s. Whatever number we pick, enough contention beats it. And a 60s × 6 = **6-minute** window to detect a genuinely hung mongod is *worse* than today for real failures.

**3. We don't lose the deep check — it moves to readiness.** The readiness probe still runs the full `mongosh` check (mongod up, reachable, authenticating, answering, correct role). The difference is what *happens* on failure: readiness pulls the pod from the Service (safe); liveness **restarts the container** (disruptive). For a stateful DB, a slow/busy member should be taken out of rotation, not killed — killing a live-but-busy PRIMARY is exactly what triggered the election + crash recovery in this bug. Also, a truly *dead* mongod (crash/OOM/panic) exits the process → the container exits → kubelet restarts it via `restartPolicy: Always`, no liveness needed.

**Where you're right:** `tcpSocket` won't restart a mongod that is *deadlocked but still holding the port open*. That's the one gap. In practice readiness removes it from the Service and the replica set heartbeats route around it, but liveness wouldn't auto-restart that specific case.

If the team wants liveness to still catch a hung-but-listening mongod, the compromise is a **lightweight command liveness** — `mongosh --eval 'db.adminCommand("ping")'` (ping only, **no** role assertion) with `timeoutSeconds: 15` and a higher `failureThreshold`. That keeps "is mongod answering" as a restart signal while dropping the election-killer and adding headroom. It's still a heavy Node.js exec though, so it's less starvation-proof than `tcpSocket` — that's the tradeoff. Happy to switch to that if you prefer it over the pure TCP check.

px-kesavan: I still have a concern on 3rd point , even the readiness probe will face the same issue isn't it ? I see you have changed the timeout from 5 to 15 , based on your 2nd point , this will still fail , so pods will be 0/1 forever as it might crosses the timeout same as the older livenessprobe

pallav-px: readiness runs the same `mongosh` exec, so under hard enough starvation it can also hit `0/1`. But that doesn't reproduce the bug, because the consequence differs:

- **Liveness fail → container restart** → primary killed → election → `ReplicaSetNoPrimary`. Outage.
- **Readiness fail → `0/1`, dropped from Service endpoints only.** No restart, no election; mongod stays alive and **stays PRIMARY, keeps serving.**

And `0/1` doesn't cut off px-backup: the headless Service sets `publishNotReadyAddresses: true` (NotReady pods stay in DNS), and RS role + client routing are driven by mongod heartbeats/topology, not k8s readiness. Live proof: under identical starvation the member went `ready=false` but **kept PRIMARY and px-backup logged 0 errors** — vs. the old probe killing the primary with 8× `ReplicaSetNoPrimary`.

"`0/1` forever" only lasts as long as the starvation — a capacity issue the `resources` change fixes (off BestEffort) — and even then it's a slow-but-serving DB, not an outage.

Net: worst case now = transiently NotReady but serving; worst case today = killed primary + no-primary window.

pallav-px: Switched to your approach for 3.1.0: **kept the `mongosh` check on both probes and just widened `timeoutSeconds` 5 → 300** (reverted the tcpSocket change). This keeps the functional "mongod up / reachable / authenticating / answering" signal you wanted, and 300s is well clear of the observed starvation latency (14–19s), so the probe no longer false-fails and kills the primary.

I've kept two caveats on the record as a follow-up (not blocking 3.1.0), since this is a stopgap rather than the final design:
1. The liveness probe still asserts replica-set role (`isWritablePrimary || secondary`), which can fail during a prolonged election regardless of the timeout.
2. A heavy `mongosh` exec on liveness stays fragile under *extreme* starvation — 300s is headroom, not a guarantee.

Opening an improvement to analyze the optimal probe/HA design and to **run a node-starvation test to confirm 300s is sufficient**. Will link it here.

pallav-px: Follow-up filed: **PB-17257** — "Optimize px-central MongoDB probe/HA design and validate under node starvation." It captures the two caveats above (role assertion in liveness, `mongosh` fragility under extreme starvation), the probe-design options, resource/QoS hardening, and a node-starvation validation of the 300s timeout under the real load profile. Targeted post-3.1.0-GA so it doesn't block this PR.

px-kesavan (approved):

pallav-px:
## Live validation of this change (300s probe timeout) — issue no longer reproducible ✅

Ran the same node-starvation repro from PB-16568 against a live cluster **with this PR's config applied** (liveness + readiness `mongosh`, `timeoutSeconds: 300`, BestEffort, chart `px-central-3.1.0`).

**Setup:** patched the live MongoDB StatefulSet to 300s timeouts, rolled out (all 3 pods healthy, RESTARTS=0), then pinned a CPU hog to the **primary's** node (mongodb-1) — same method that killed the primary with the shipped 5s config.

**Probe latency under starvation:**
```
run 1: 19,693 ms   run 2: 21,292 ms   run 3: 13,399 ms   [all PASS]
old budget 5,000 ms  →  would fail/kill
new budget 300,000 ms →  passes
```

**Held starvation ~4 min (old config killed the primary in ~50s):**
| Check | Result |
|---|---|
| mongodb-1 (primary) RESTARTS | **0** |
| Liveness `Killing`/`failed liveness` events during window | **none** (only event was the earlier rollout) |
| Replica-set primary | **retained** — mongodb-1 still `PRIMARY`, all members `health=1` |
| px-backup `ReplicaSetNoPrimary` / server-selection errors | **0** |

The `mongosh` exec ran slow (13–21s) but stayed well under 300s, so the probe passed, the primary was never restarted, no election fired, and px-backup stayed connected the whole time. **The flake is not reproducible with this change.**

Cluster restored to as-found afterward (timeouts reverted, hog + namespace removed). Residual caveats (role-assertion-during-election, extreme-starvation headroom) remain tracked in PB-17257.

PR #970, opened 2026-08-18
PB-11273: show SMTP Configuration Name in alert emails

t-nidhin:
## Summary
- Alertmanager exposes the matched receiver's name as `.Receiver` in notification templates. Render it (stripped of any namespace/CR-name prefix prometheus-operator adds when merging AlertmanagerConfig CRs) in `pxc_template.tmpl` so customers running multiple testbeds on the same SMTP server can tell which environment an alert email came from.
- Pairs with the px-backup change (separate PR) making the receiver identity carry the SMTP Configuration Name instead of the recipient's name.

## Testing summary

**Live cluster test - core fix**: built and deployed this branch, created a SMTP receiver + recipient via the API, confirmed the generated `AlertmanagerConfig` CR and prometheus-operator's merged Alertmanager config carry the correct Configuration Name (not the old recipient-derived name), and fired an alert through Alertmanager.

**Live cluster test - migration**: manually forced an existing CR back to the pre-fix stale naming scheme, reset the migration's tracking doc in MongoDB, restarted the pod, and confirmed the migration goroutine rebuilt the CR's receiver name and all 3 route receiver values back to the correct SMTP Configuration Name on that restart.

Screenshot 2026-08-18 at 2 08 09 PM


t-nidhin:
## Testing summary

**Render check**: `helm template` with `pxbackup.enabled=true` / `deployDedicatedMonitoringSystem=true` renders cleanly; `Configuration:` now sits as the first row in "Alert details" for every alert type (moved there per feedback - not a standalone line above the alert loop), captured once via `$configName` before `range .Alerts` since `.Receiver` is a top-level template field.

**Live cluster test** (sys1): patched the deployed `pxc_template.tmpl` secret with this change, verified prometheus-operator's merged Alertmanager config mangles the receiver name as `<namespace>/<CR-name>/<SMTP-config-name>` (confirming `reReplaceAll ".*/" "" .Receiver` correctly strips down to just the config name), then fired a synthetic alert through Alertmanager's real notify pipeline - it routed to the right receiver and reached the email-send stage. Cluster's secret reverted to its original content afterward.

Pairs with the px-backup change (pure-px/px-backup#2847) making the receiver identity itself carry the SMTP Configuration Name.

yassharma-px (approved):

kgarg-px (approved):
LGTM

